PR #270, opened 2015-02-01
Adds google drive folder icons

wa4557:
Since we are hopefully getting an official goolge drive client for linux (fingers crossed: http://www.webupd8.org/2015/01/official-google-drive-linux-client.html), I thought it is necessary to have the proper places icons...

### DIFF
--- a/Numix/128x128/places/folder-gdrive.svg
+++ b/Numix/128x128/places/folder-gdrive.svg
@@ -1,0 +1,207 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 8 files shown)


Foggalong:
Seems like a cool addition, but do we how do we know what the icon name will be?

bil-elmoussaoui:
if so , you can support more folders like steam folder or virtualbox mks foldr?

Foggalong:
Plan is to add a few more folder types, so feel free to request some in separate issues. The above question was more out of curiosity than a demand

wa4557:
Nope, we don't know yet, but symlinking should be the smallest problem :)

Foggalong:
We decided to go with a slightly different symbol and had to set it all up in the [Numix Folders](https://github.com/numixproject/numix-folders) repo rather than here. However, I'm gonna merge this in and then make another pull to revert it so that you get credit for your hard work. Seriously - thank you :) We wouldn't have done a Google Drive icon without you!

Issue #248, opened 2014-12-04
Ubuntu notification icons - volume


dirtydancing:
Currently, there are no Numix Base icons for the notifictions in Ubuntu (NotifyOSD), cf. overview at https://wiki.ubuntu.com/NotifyOSD#Icon. I would like to suggest to add these.

These icons per default are of size 48. In Ubuntu (14.10), they are located per default in `usr/share/notify-osd/icons/gnome/scalable/status`. So I would suggest to create a new folder in Numix Base (and correspondingly in index.theme) named `48x48/notifications`.

A good place to start this would be the volume notification icons. I have tested this with the current Numix Base 22x22 .svg volume icons (renamed and added to an extra folder `22x22/notifications` for testing purposes), and this has worked just fine.

The relevant icons here are
`notification-audio-volume-high.svg`
`notification-audio-volume-low.svg`
`notification-audio-volume-medium.svg`
`notification-audio-volume-muted.svg`
`notification-audio-volume-off.svg`

Volume icon overview screenshot:
![notification-icons_volume](https://cloud.githubusercontent.com/assets/7164227/5299056/307d2710-7bc3-11e4-849f-6b52798656dd.png)

Comparative screenshots of the Ubuntu default and the Numix notification when changing the volume:

Ubuntu default as of present:
![notification_default](https://cloud.githubusercontent.com/assets/7164227/5298996/bc731fa0-7bc2-11e4-9159-7940f95c69e5.png)

With Numix volume notification icon:
![notification_numix](https://cloud.githubusercontent.com/assets/7164227/5299005/c10634a8-7bc2-11e4-83b7-b02c29a64dcf.png)

dirtydancing:
Mobile broadband added: https://github.com/numixproject/numix-icon-theme/issues/287

wa4557:
Aaand fixed :)

Issue #271, opened 2015-02-01
Keyboard indicator 


bil-elmoussaoui:
Hello, i've a little problem,
i'm using Numix Icon Theme on my Elementary OS 0.3 Beta 1 
![capture du 2015-02-01 22 16 51](https://cloud.githubusercontent.com/assets/7660997/5993904/228ccd1a-aa60-11e4-8fcb-eae300174ca7.png)
this is my keyboard indicator for the french language ; when i add an other language like english :  and i chose this one the icon became like this ? but only for the english 
![capture du 2015-02-01 22 17 07](https://cloud.githubusercontent.com/assets/7660997/5993905/3d22a97e-aa60-11e4-90a3-e44b3ac7ac1d.png)
i added one more language to test it (arabic) and i dont have this problem? 
![capture du 2015-02-01 22 17 30](https://cloud.githubusercontent.com/assets/7660997/5993907/587824ec-aa60-11e4-9a26-9cf9186d466b.png)

I dont know if i have to give you more informations? 
Thanks

dirtydancing:
Currently, Numix base comprises only a couple of indicator-keyboard icons, i.a. for EN. As elementary OS Freya (Beta) is using the canvas size 24x24 in the wingpanel, for EN the current Numix base icon is used. For the other languages in your screenshots above, as these icons are white and smaller, to me this looks like they are taken from elementary icon theme (which seems to be inherited in your settings).

I am not sure why the "EN" looks so greyed out in Numix base in size 24, this is different e.g. in size 22. In the following screenshots you can see which indicator-keyboard icons are currently in Numix base, and that the colours differ inbetween size 24x24 (e.g. elementary OS Freya) and 22x22 (e.g. Ubuntu).

`Numix/24x24/status` (EN marked, somewhat "greyed out", like in your screenshot):

![numix_24x24_status](https://cloud.githubusercontent.com/assets/7164227/6433349/b3b91862-c06c-11e4-8df4-79538ca5b6e4.png)

`Numix/22x22/status` (EN marked):

![numix_22x22_status](https://cloud.githubusercontent.com/assets/7164227/6433354/d6ba87d8-c06c-11e4-880b-87d0291857a3.png)

dirtydancing:
In addition, the indicator-keyboard icons in Numix base are also "greyed out" in size 16x16, just as in 24x24.

`Numix/16x16/status` (EN marked):

[...]

Issue #255, opened 2014-12-08
Ubuntu notification icons - battery alerts


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/issues/248, https://github.com/numixproject/numix-icon-theme/issues/251, https://github.com/numixproject/numix-icon-theme/issues/252 and https://github.com/numixproject/numix-icon-theme/issues/253, this issue concludes the mini-series of Ubuntu notification icons (NotifyOSD) for the time being.

This has been tested in current Ubuntu 14.10/Unity with the default settings. In practice, the battery alerts turned out to be apparently a mixture of time and percentage alerts, 5 alerts altogether, although the value "use-time-for-policy" in "org.gnome.settings-daemon.plugins.power" was default on, and only a trilogy of (1) low, (2) critical and (3) action was to be expected.

The present issue is a tricky one. But looking back at all the previous Numix Base battery icon issues, I figure it would not really be the battery icon if this was a trivial and straightforward process :-).

The relevant icon names for the battery notifications are:
`notification-battery-020.svg`
`notification-battery-low.svg` (symlink with target notification-battery-020.svg)
`notification-battery-000.svg`
`notification-battery-empty.svg` (symlink with target notification-battery-000.svg)
`battery-caution-symbolic.svg` (symlink with target notification-battery-000.svg)

Screenshot with an overview of the relevant icons:
![notifications-battery-alerts_overview](https://cloud.githubusercontent.com/assets/7164227/5337264/dc95a3aa-7ec6-11e4-801f-cf78f86a579a.png)

1st alert: battery low (time 20 min): `notification-battery-low.svg`

Comparative screenshots:

Current
![1a_notification-battery-low-time_current](https://cloud.githubusercontent.com/assets/7164227/5337280/f0157ac2-7ec6-11e4-9360-d89ced231216.png)

With Numix Base `notification-battery-low.svg`
![1b_notification-battery-low-time_numix](https://cloud.githubusercontent.com/assets/7164227/5337288/fc7a6656-7ec6-11e4-98d1-0a3f5ad0c009.png)

2nd alert: battery low (percentage 10%): `notification-battery-020.svg`

Comparative screenshots:

Current
![2a_notification-battery-low-percentage_current](https://cloud.githubusercontent.com/assets/7164227/5337294/103c2a80-7ec7-11e4-8c58-ed5f5058cac8.png)

With Numix Base `notification-battery-020.svg`
[...]

Issue #253, opened 2014-12-06
Ubuntu notification icons - wired


dirtydancing:
In continuation esp. of https://github.com/numixproject/numix-icon-theme/issues/252 for wireless, here are the corresponding notification icons for wired:

The two relevant notification icon names for wired are:

`notification-network-ethernet-disconnected.svg`
`nm-device-wired.svg`

Screenshot comparison for wired disconnected in current default and with Numix notification icon: Suggested here is the same icon for the notification that is also showing in the panel. Alternatively, to make this icon more distinct, a "wired" icon with a red or white cross (for disconnected) in it could be created, similar to the current default icon approach.

Notification wired disconnected current:
![notification-wired-disconnected_current](https://cloud.githubusercontent.com/assets/7164227/5328642/b17f478c-7d88-11e4-9a61-44b21258c2f6.png)

Notification wired disconnected with Numix Base notification icon:
![notification-wired-disconnected_numix](https://cloud.githubusercontent.com/assets/7164227/5328643/b84f0d54-7d88-11e4-9f62-63e2628faac8.png)

The "notification" icon for wired connected, in practice, actually is `nm-device-wired`.

Now, currently in Numix Base, there are two kinds of `nm-device-wired` icons:

First, in `status` with the two arrows:
![nm-device-wired_status](https://cloud.githubusercontent.com/assets/7164227/5328649/d7a3459e-7d88-11e4-940a-28deb530f5b4.png)

and second, in `devices`, looking like this:
![nm-device-wired_devices](https://cloud.githubusercontent.com/assets/7164227/5328652/e960e1ce-7d88-11e4-9d41-7bda10445c8a.png)

Overview screenshot of `nm-device-wired.svg` in current Numix Base:
![nm-device-wired_overview](https://cloud.githubusercontent.com/assets/7164227/5328656/0500e2e4-7d89-11e4-91ba-0ad3520a53d6.png)

This might be a good opportunity to make these icons consistent, and I would suggest to adopt the design with the two arrows also for the `devices` icons. That is unless there is a reason for the quite divergent icon designs inbetween `status` and `devices`, of course.

Screenshot comparison for wired connected currently (icon apparently picked from a Numix Base `devices` folder) and for testing purposes with `nm-device-wired.svg` `status` icon:

Notification wired connected current:
[...]

PR #283, opened 2015-03-07
add all languages

wa4557:
Adds 22/status indicator-keyboard for all languages. I used #dcdcdc instead of #dbdbdb as base color since it is the one used most of the times


wa4557:
I deleted the keyboard indicators from 16x16 and 24x24 since they were wrong anyway and are useless imo. Prove me wrong and I'll add proper sized icons there as well

dirtydancing:
Please cf. my comment here https://github.com/numixproject/numix-icon-theme/issues/271#issuecomment-77699560

dirtydancing:
@wa4557 Great to see progress with the Ubuntu notification icons! I am in the course of testing the  notification icons that you added. This is quite tricky and will take some more time. I will have feedback.

From my perspective, the keyboard indicator icons are ready to be merged, but the notification icons are not. Is it possible/sensible to separate these two? I am not familiar with how this is handled on GitHub.

wa4557:
Probably, but I suppose it will take some time till it is merged so most definitely not worth it...

dirtydancing:
OK, here is a first round of feedback regarding the (Ubuntu) notification icons:

1) Icon (canvas) size: These should be 48x48 icons, located in `48x48/notifications`. Reasons:

First, the Ubuntu notification icons in Notify OSD were originally designed in 48x48, and experience shows that, although these are scalable .svg files, it pays off to have them in the relevant size.

Second, the notification icons should be separated from the Ubuntu panel status icons. The notification icons are a set of their own, and they do not primarily go for size 22x22 icons like the Ubuntu panel. I checked e.g. with `nm-device-wwan.svg`: in the presence of icon sizes 16x16, 22x22 and 24x24, the Ubuntu notification bubble will go for the 24x24 icon first, only after that for size 22x22, and finally for size 16x16.

Third, by having a separate section of size 48x48 icons, the notification icons become independent of potential design changes in the panel (currently, there are several symlinks to panel icons).

Fourth, currently the `scalable` directory is interfering with the fixed size directories (specifically for `battery-caution-symbolic.svg`, cf. https://github.com/numixproject/numix-icon-theme/issues/255. I will not go into further detail here, if you have a lot of time and then some more time, check https://github.com/numixproject/numix-icon-theme/issues/107 :-) The point here is that possibly it will make sense to put the notification icons into Numix-Circle instead of Numix base, but that will require further testing with 48x48 icons in Numix base first and also (potentially) "collision" testing in Numix-Circle second. Also for this scenario, it makes sense to have a "separate" folder for the notification icons that is independent of other icons in Numix base.

Fifth, even in Numix base, there are some some differences between different connection type icons that the notification bubble is using currently (e.g. there is no `nm-device-wireless.svg` in sizes 16, 22 or 24, only in 48), so here is an additional reason for a separate `48x48/notifications` folder that can be maintained on its own.

This addition of Ubuntu notification icons WILL be noticed by users. I have not seen any non-Numix notification icons on my machine for months, and seeing some now is visually quite irritating. Then again, I have had the luxury of creating a custom icon theme with a couple of modifications and putting that before Numix-Circle in the first place of the icon theme inheritance line. It is much more tricky when the notification icons are supposed to work as good as possible in Numix base/Numix-Circle, and at the same time not cause any "breakage" anywhere else.

Convinced of the `48x48/notifications` folder and 48x48 icon (canvas) size? :-)

Such an additional folder would then also have to be listed index.theme of Numix base (I can see that you currently added `22x22/notifications` to index.theme).

2) Icon (actual icon) size

The canvas size being 48x48, the next question is how large the actual icons should be on that canvas. I would recommend to scale these up from the current (if present) 24x24 icons. And I would recommend to keep the actual icon size on the 48x48 canvas consistent throughout all the notification icons.

3) Icon design

I have not got into detailed drop shadow presence (or not)/consistency or the like so far.

Here, I would like to recommend to pay special attention to the different "disconnection" notification icons for different connection types. The notification icons should pick up the differentiation that the original icons offer and display the specific connection type that just got disconnected. As reported in https://github.com/numixproject/numix-icon-theme/issues/252, https://github.com/numixproject/numix-icon-theme/issues/253 and (NEW) https://github.com/numixproject/numix-icon-theme/issues/287, I am referring to these three icons:

disconnected

This would mean to take the respective Numix "connected" icon and add some sort of red "x" to it (again, this shows that the notification icons stand on their own, cf. above).

The next question then would be what colour the red should be, and this leads to my other new issue report over here: https://github.com/numixproject/numix-icon-theme/issues/286. This is a similar situation as with the light grey colour inconsistency. The two warning reds in Numix basically look identical, and I figure for the time being just pick one and go along with it :-)

Please let me know if anything is unclear so far or if you think anything should be done differently. If not, could you make the according adjustments? I will then have another look at it.

dirtydancing:
Addition concerning the icon design: in continuation of https://github.com/numixproject/numix-icon-theme/issues/253 I would like to point out that there is a third icon for `nm-device-wired` in Numix base, which is located in `22x22/devices`. I would suggest to use this design for the notification icons for wired, because it is the most specific, and also this is closest to the original notification icons, cf. screenshot above.

nm-device-wired_1

nm-device-wired_2

As a side note, the same icon shape is used for `network-wired-symbolic.svg` in `scalable/status` (size 16x16, employed most notably in GNOME Shell panels):

network-wired-symbolic

dirtydancing:
Also, for the network notifications (wireless, wired, mobile broadband), I would suggest to add the icons for connected and disconnected with the icon names `...-connected` and `...-disconnected`.

Well, to be precise, in `notify-osd/icons/gnome/scalable/status`, the icon name '...-connected` is only used for ethernet-connected, not for wireless-"connected" or for gsm-"connected". Possibly those icon names for wireless and for gsm should be different. My point here is to have a set of two corresponding "notification"-named icons for the network connections (for "connected" and for "disconnected", as a base line.

Currently, as reported, for all three the "notification"-named icon is used (checked in Ubuntu 14.10) for disconnected, but in contrast for connected, the icon `nm-device-...` is used. That could be added in an additional step e.g. as a symlink to the corresponding "connected"-"notification"-named icon. Anyways, it would remain to be seen whether such icon is picked up, and whether there would be a conflict with already existing `nm-device-...` icons.

Alas, I was just trying to reduce the confusion around the notification icons, but I am afraid I have just added some more confusion instead :-)

wa4557:
Yeah cool, I'll make the adjustments as soon as possible, give me a day or two. Sounds reasonable!

dirtydancing:
Great!

Also one more addition as to the icon (actual icon) size: I noticed that the original notification icons appear to use an actual icon size of 48x48 on a 48x48 canvas (they have a kind of "shade" all arond them, though). And e.g if you look at the comparison of the original and the (test) Numix icon for volume notification, https://github.com/numixproject/numix-icon-theme/issues/248, the Numix one looks somewhat small.

notification-audio-volume-high

notification-network-ethernet-disconnected

So I would suggest to create the Numix notification icons in `48x48/notifications` on a 48x48 canvas and with an actual icon size that resembles that of the original notification icons. This will be larger relative to the canvas size than e.g. the icons in `22x22/status` or `24x24/status` in Numix base, or at least this will definitely be larger relative to the canvas size than the icons in `24x24/status` (which I have recommended as a reference for the actual icon size above).

As a visual comparison, current screenshot from `24x24/status`

audio-volume-high-panel

dirtydancing:
Actually, when I cross-checked with the actual icon size in 22x22/status, the actual icon size looked comparable (just not from current 24x24). So I figure scaling up from 22x22 would be fine indeed.

As a visual comparison, current screenshot from `22x22/status`

audio-volume-high-panel_22x22 png

dirtydancing:
Icon design (colour):

One more aspect as to the notification icon colour, here specifically for volume and brightness: As is visible in the screenshots in https://github.com/numixproject/numix-icon-theme/issues/248, for volume and brightness there is a bar indicating the level. This bar has a somewhat different style, and it is a little darker than the Numix icon. As far as I can tell, this bar is not themeable (?). It might be worth checking if the Numix notification icons (for volume and brightness) make a more constistent impression in the notification bubble and also overall when their colour is made a little darker grey (?). Just an idea.

The good thing here is that the notification icons for volume and brightness work properly in their "notification-" naming and that thus there will be no "icon collision" with other volume (status) icons whatsoever. Arguably, these are the two most visible notification icons (esp. the volume notification icon), so this icon colour question is worth considering.

dirtydancing:
Icon design: Please also cf. the pull request for battery-000.svg at https://github.com/numixproject/numix-icon-theme/pull/290. This is also relevant for the battery notification icons, cf. https://github.com/numixproject/numix-icon-theme/issues/255.

dirtydancing:
In addition, please also cf. the pull request for removing nm-device-wireless.svg from Numix-**Circle** https://github.com/numixproject/numix-icon-theme-circle/pull/1690, also with relevance for the displayed notification icon for wireless "connected": https://github.com/numixproject/numix-icon-theme-circle/issues/1361, https://github.com/numixproject/numix-icon-theme/issues/252

wa4557:
It is not themeable, but it can be modified, by modifiying the source code of notify-osd. I wouldn''t  change the colors, and use the standard: #dcdcdc

unbenannt

When we're finished with this whole nofification issues I'm  thinking about creating a ppa for a numix-notify-osd or sth. similar. But I've never done sth like this, so I have to see how this is done...

dirtydancing:
Interesting! And indeed, with "themeable" I was referring to themeable via the icon theme (to have this in one place). And agreed, Numix standard colours make sense here. Thin line here walking the balance between Numix standards and a pragmatic approach :-)

PPA for numix-notify-osd, wow! On the other hand, iirc correctly, there is already a PPA for modifying the notify-osd notification bubbles, maybe it would make sense to add the possibility to modify the "bars" there? But then again, that would still require more instruction to suit the Numix design...

wa4557:
Ok I've added the notification icons for volume and brightness in 48x48/notifications. And I removed the drop-shadow (The shadow doesn't make a lot of sense, imo). Adding battery etc. is more complicated

dirtydancing:
**1)** In continuation of the conversation over at https://github.com/numixproject/numix-icon-theme/issues/252, I would like to come back to `index.theme`:

Right now, `48x48/notifications` appears to be working just fine, but it is not listed in the alphabetical order. I would suggest to have this right after `48x48/mimetypes` (above and below). This might seem marginal, but in my experience, even the slightest potential error in `index.theme` should be avoided.

If you are interested in an example of this, I recommend e.g. https://github.com/numixproject/numix-icon-theme/issues/165 :-)

**2)** Now that wireless and wired are working, for the connections this leaves mobile broadband, cf. https://github.com/numixproject/numix-icon-theme/issues/287. Do you think you can add these two icons as well? This would then be a consistent set of connection notification icons.

I tested the mobile broadband on a friend's machine, so I would not be able to test this right away. This should not slow us down, though, and one possible way to await testing confirmation and still have these icons available, would be e.g. to add `_proposed` at the end of the two icon names.

Seeing how good wireless and wired are working, there is a high probability that mobile broadband will just work as well.

**3)** I saw your message about the battery icons. This might be the trickiest part of them all. I will check this asap, but this might take some more in-depth testing.

Did you get all the correct battery icons for all the five states that I reported at https://github.com/numixproject/numix-icon-theme/issues/255? I am particularly interested in the 4th and 5th alerts, that pick up `battery-caution-symbolic.svg`, because there could be a possible collision with the `scalable/status` directory.

wa4557:
1. i'll push the new index.theme in a few hours 
2. Working on them right now :)

3. I Saw all warnings and they were ccorrect, so at least for Ubuntu it seems to work. If the battery-caution-symbolic.svg interferes with other distros idk

dirtydancing:
1. Great!

2. Fantastic!

3. Unfortunately, I can NOT confirm this for Ubuntu (14.10). On my machine, for warnings No. 4 and 5 ("Critical condition...") `battery-caution-symbolic.svg` from `scalable/status` is displayed, not from `48x48/notifications`. 

This is the icon (symlink) in `48x48/notifications`:
notifications

In contrast, this is the icon (16x16) in `scalable/status` (darker grey, larger in relation to the canvas):
scalable

All the warnings are shown, but the notification icon for the last two warnings is the wrong one (sometimes the exact sequence of the warnings may differ). Here are two comparative screenshots, first and fourth warning notification. In the first one, the correct notification icon is showing, in the fourth one it is `battery-caution-symbolic.svg` from `scalable/status`:

First battery warning notification (correct icon):
warning-1

Fourth battery warning notification (wrong icon):
warning-4

Sorry for asking again: for the final battery warnings, is the red battery icon from `48x48/notifications` displayed in the notification bubble on your machine?

dirtydancing:
To me, `battery-caution-symbolic.svg` appears to be the wrong icon for the notification bubble to pick up, but I would not even know where to begin to look for a possible bug in Ubuntu/Upstream for that.

And anyways, also `battery-caution-symbolic.svg` in `scalable/status` is a warning (orange) battery icon, so it does fulfill its purpose, although a somewhat inconsistent icon is displayed (at least on my machine).

I am saying this because the only way I can think of to "work around" this inconsistency is to place the directory `48x48/notifications` in Numix-**Circle**, so these icons (including `battery-caution-symbolic.svg`) would get looked at before Numix base with its `scalable/status` directory even comes into play.

But with all the other notification icons (including all the other battery notification icons) so far working from Numix base, these icons should not be placed in Numix-Circle, because that might break all sorts of things in other distributions (or not, but that would need to be checked). 

In addition, these kind of icons belong into Numix base, not into Numix-Circle, because they are not application icons. An exception to this should only be made when it is absolutely necessary (such as is the case e.g. for the **app** panel status icons).

And further, it is not an option to e.g. remove `battery-caution-symbolic.svg` from `scalable/status`, because this very icon is needed for all GNOME distributions, cf. https://github.com/numixproject/numix-icon-theme/issues/107.

But before we come to any conclusions here, let's first see whether you can confirm this situation on your machine.

wa4557:
I'm afraid you're right. I can reproduce that now. Wasn't looking at the right notifications before... Well that's a pity. But since that's a ubuntu/unity bug there's not much we can do about it.

dirtydancing:
Good, thanks, now at least we know that this is reproducible. Whether this is a bug or not and whereever exactly this is originating from, the question remains how to deal with this from Numix's side.

As I already noted, in the end also `battery-caution-symbolic.svg` from `scalable/status` is OK as an icon, as it fulfills the warning function. And before that one even shows up, several other battery notification warnings with proper icons from `48x48/notifications`will be displayed, so this is not all that bad after all.

As a result of this, and also because `battery-caution-symbolic.svg` does not really belong into the fixed size directory `48x48/notifications` and is an anomaly among all the other notification icon names (only one that is "symbolic"-named), I would probably just remove this symlink from `48x48/notifications`. That way, such a "symbolic"-named symlink could also not cause any harm in other distributions that are making more use of "symbolic"-named icons. Then again, the directory `48x48/notifications` is rather special, and should not be looked into too much for other purposes than, well, notification icons :-)

Or do you think that this symlink should stay in `48x48/notifications`, just in case? There might be an argument for that as well. Possibly for those who might be deleting the "symbolic"-named battery icons in Ubuntu (e.g. because that has an effect on proper display scaling in Ubuntu, again cf. https://github.com/numixproject/numix-icon-theme/issues/107). At least when there is no `battery-caution-symbolic.svg` icon in `scalable/status`, then the "proper" one in `48x48/notifications` should be displayed (again, this would need to be tested).

What do you make of this?

dirtydancing:
Oh, I am presuming with the question above that the directory `48x48/notifications` is placed in Numix base (and not in Numix-Circle), which is currently the case, as this is a pull request for Numix base, and which I recommend, cf. above.

wa4557:
I think it is better to keep it, if someone wants to manually delete the one in the scalable directory. For me it seems that when this icon is needed the one from the scalable directory is used -> the one in `48x48/notifications` won't do any harm.

Let's keep the notification icons in the Base-directory, it doesn't make sense in the Circle repository as you already stated above

Ok. I think we are finished here. Seems to me we did as good as possible

Foggalong:
It's just good practice to keep each individual new feature, fix, or enhancement in a separate pull request. If you don't, as has happened here, it becomes nigh-on-impossible to check that each part is correct and up to spec. Really sorry that I didn't catch this sooner, but just I can't merge this pull request :(

Foggalong:
That said, I'd be more than happy to get your contributions in there because it looks like you've been working on some good stuff here that fixes a lot of outstanding issues. It just needs to be better organized

dirtydancing:
@wa4557 Again, I am not familiar with how to handle this kind of situation on GitHub. These are all useful additions indeed, and this appears to be a matter of splitting them up into separate pull requests.

On what we were discussing above: I agree with you that it is better to keep the symlink `battery-caution-symbolic.svg` in `48x48/notifications`. This way, the "notification icons" are completed in one place. 

I checked and put the `48x48/notifications` folder into a custom icon theme that does not contain a `scalable/status` directory, and now all the battery notfication warnings were displayed correctly. Good!

Furthermore, I checked with Ubuntu GNOME 14.10 (default GNOME Shell 3.12.2) with your latest commit from above, and the battery status icon `battery-caution-symbolic.svg` was properly displayed in the panel from the `scalable/status` directory. This means that `battery-caution-symbolic.svg` in `48x48/notifications` does not interfere with the battery status icons in GNOME shell. Good!

wa4557:
@Foggalong, Fine I understand. I'll split them up as good as possible. But I think I can't do this for the bug-fixes I pushed for the `numix-folder` repo. 

First they are mostly minor bugs, that change only one or two files, and secondly the different commits do interfere with one another, so pushing each bug-fix seperately will result in conflicts imo. Nevertheless I'll split what I can...

Foggalong:
In the case of conflicts it's best to submit them one after the other rather than simultaneously. Thanks for the understanding :) Just to let you know (nothings confirmed yet) but it's possible we're going to have a redesign of the status icons to fix some problems we've had with the light theme and transparent pannels. I'll keep you posted!

dirtydancing:
I can now also confirm that the notification icons for mobile broadband are working in Ubuntu 14.10.

With this, all the suggested notification icons have been tested to be working. Great!

I have already incorporated them on my machine, and everything is working smoothly and consistently.

And in case that there will be a redesign of the status icons in Numix base, no problem, the 48x48 notification icons could be adapted accordingly. The important thing here is that with this, a functioning "notification baseline" is established. Fine tuning can always be done, also regarding the icon names, in case that for whatever reason these should change.

This alone fixes https://github.com/numixproject/numix-icon-theme/issues/248, https://github.com/numixproject/numix-icon-theme/issues/251, https://github.com/numixproject/numix-icon-theme/issues/252, https://github.com/numixproject/numix-icon-theme/issues/253, https://github.com/numixproject/numix-icon-theme/issues/255 and https://github.com/numixproject/numix-icon-theme/issues/287.

Here is an overview screenshot of the suggested size 48x48 icons/symlinks in the new folder `48x48/notifications` (unselected and also selected for better visual contrast):

notifications

notifications_selected

wa4557:
@Foggalong since you're currently working on a new set of status icons (I shamelessly looked at your testing repository :stuck_out_tongue: ), all the fixes I proposed (pretty much all of my pull request) are temporary. That being said, are these changes reasonable nevertheless (and should I continue pushing fixes that addresses these inconsistencies), or should I stop posting them since it is a waste of time for me and esp. for you?

Issue #252, opened 2014-12-06
Ubuntu notification icons - wireless


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/issues/248 and https://github.com/numixproject/numix-icon-theme/issues/251, now on to the notification icons for wireless:

The two relevant notification icon names here are:

`notification-network-wireless-disconnected.svg`
`nm-device-wireless.svg`

Screenshot comparison for wireless disconnected in current default and with Numix notification icon:

Notification wireless disconnected current:
![notification_wireless_disconnected_current](https://cloud.githubusercontent.com/assets/7164227/5327690/a1fcb2d6-7d5b-11e4-81e0-7a8ceec81f74.png)

Notification wireless disconnected with Numix Base notification icon:
![notification_wireless_disconnected_numix_base](https://cloud.githubusercontent.com/assets/7164227/5327693/c213be20-7d5b-11e4-8b27-12b25636fb6d.png)

The "notification" icon for wireless connected actually is `nm-device-wireless`. This would be a regular Numix Base status icon e.g. in `22x22/status`.

This is a longstanding limitation (feature?) of NotifyOSD/Upstream (no actual signal strength displayed in the Ubuntu notification bubble), as can be seen in these two bug reports:

https://bugs.launchpad.net/hundredpapercuts/+bug/387626
https://bugzilla.gnome.org/show_bug.cgi?id=686357

Still, in my view, this would not be a case of Numix catering to a bug in another project. Rather, currently `nm-device-wireless.svg` actually is an icon also in Numix-Circle, although this icon belongs to Numix Base, cf. https://github.com/numixproject/numix-icon-theme-circle/issues/1361 over at Numix-Circle. So having this icon only in Numix Base, just like e.g. `nm-device-wired.svg`, would automatically lead to it being displayed correctly in the Ubuntu notification bubble.

Screenshot comparison for wireless connected currently and with `nm-device-wireless.svg` status icon (for testing purposes, I renamed an existing 22x22 Numix icon accordingly in a somewhat different icon theme structure, so this is only for illustrating):

Notification wireless connected current:
![notification_wireless_connected_current](https://cloud.githubusercontent.com/assets/7164227/5327702/431cf072-7d5c-11e4-8760-a153c367f4f6.png)

Notification wireless connected with Numix Base `nm-device-wireless.svg` icon:
[...]

PR #298, opened 2015-03-17
notification icons

wa4557:
Pull request for wireless, wired, battery, wwan, audio and brightness notification icons. Closes 5 bugs


dirtydancing:
In addition to the comments on the notification icons in https://github.com/numixproject/numix-icon-theme/pull/283 and https://github.com/numixproject/numix-icon-theme/issues/252, I would like to point out that these notification icons will get even more relevant in Ubuntu 15.04.

This is because in Ubuntu 15.04 (currently in development), the notify-osd-icons are placed in Humanity rather than in (until now) gnome icon theme, cf. https://launchpad.net/ubuntu/+source/notify-osd-icons/0.8+15.04.20150202-0ubuntu1 and http://packages.ubuntu.com/de/vivid/all/notify-osd-icons/filelist.

As Numix inherits i.a. gnome, but not Humanity, this means that the notify-osd-icons will not get picked up anymore. I check this on Ubuntu 15.04 in a VM with network wired, and indeed, when disconnected, the notification bubble showed without any icon, but for connected (as the "nm-device..."-named icon is picked up rather than a "notification"-named icon), the same (inconsistent) icon showed in the notification bubble as currently in Ubuntu 14.10, cf. https://github.com/numixproject/numix-icon-theme/issues/253.

I checked with the notification icons proposed here, and for network wired all was picked up correctly, just like already in Ubuntu 14.10.

wa4557:
@Foggalong should I update these icons in order to be consistent with the new status icons?

wa4557:
Closing this one and opening a new one that has no conflicts

Issue #240, opened 2014-11-12
user-trash-full-symbolic.svg not showing in Ubuntu 14.10 with default Ambiance GTK theme


dirtydancing:
I have encountered a rather strange phenomenon in default Nautilus File Manager (3.10.1-0ubuntu15) in Ubuntu 14.10 with default Ambiance GTK theme (same for Radiance GTK theme): when the rubbish bin is full, the symbolic red Numix icon is very hard to see, as a light grey colour is displayed on a light grey background. This does not happen with the Numix GTK theme, though.

Screenshot with Numix GTK theme:

![trash1](https://cloud.githubusercontent.com/assets/7164227/5007772/a573aef6-6a55-11e4-8344-24d1dcafc688.png)

Screenshot with Ambiance GTK theme (rubbish bin marked so this can be seen at all):

![trash2](https://cloud.githubusercontent.com/assets/7164227/5007776/acc1060e-6a55-11e4-8af0-f7bba649f170.png)

On the other hand, with Ambiance GTK theme, the user-trash-full-symbolic.svg is showing correctly for other icon themes, such as for the default Ubuntu icon theme and for Moka/Faba icon theme.

So this might not be a bug anywhere, as the respective "defaults" are working with each other. Still, this is the first time that I am noticing that the Numix icon theme is not playing well with the Ubuntu default Ambiance GTK theme, so I am filing this issue here.

Possibly there is an "easy" way that this can be addressed from Numix's side?

I have found a "workaround" by modifying the icon in question: e.g., if it looks like this, it is displayed with Ambiance GTK theme as well. So apparently there can be colours in the icons, and this seems to relate to the specific red colour that the Numix icon is currently using (?):

![trash3b](https://cloud.githubusercontent.com/assets/7164227/5007806/eedb4ee6-6a55-11e4-9807-058335cccee0.png)

![trash3a](https://cloud.githubusercontent.com/assets/7164227/5007800/ea24af64-6a55-11e4-968a-d41084c70ade.png)

dirtydancing:
Bug report filed against Ambiance GTK theme (ubuntu-themes): https://bugs.launchpad.net/ubuntu/+source/ubuntu-themes/+bug/1426774

dirtydancing:
Quick update: by now, this problem has been qualified as a papercut bug in Ubuntu, cf. https://bugs.launchpad.net/ubuntu/+source/ubuntu-themes/+bug/1426774

[...]

PR #303, opened 2015-03-20
Added user-trash-full-symbolic-AmbianceGTK.svg

dirtydancing:
Added `user-trash-full-symbolic-AmbianceGTK.svg`  to `/scalable/places`. Temporarily addresses https://github.com/numixproject/numix-icon-theme/issues/240 from Numix's side.

In continuation of https://github.com/numixproject/numix-icon-theme/issues/240, I would like to suggest to add this "workaround" icon to Numix base. This has worked just fine for me for a couple of months now with Ubuntu default Ambiance GTK theme.

workaround-icon

Bug report against Ambiance GTK theme: https://bugs.launchpad.net/ubuntu/+source/ubuntu-themes/+bug/1426774

Here is a screenshot of the result in Ubuntu 14.10 with Ambiance GTK theme (after renaming this "workaround" icon to `user-trash-full-symbolic.svg`):

user-trash-full-symbolic-ambiancegtk

In contrast, this is what is displayed with current Numix icon theme in combination with Ambiance GTK theme (same problem in Ubuntu 15.04 daily, currently in development):

06ef43e8-c118-11e4-89bd-0201742ef55b

Temporarily adding this "workaround" icon would give the user the opportunity to manually change the icon names and use this visible and functional icon in combination with Ambiance GTK theme.

### DIFF
--- a/Numix/scalable/places/user-trash-full-symbolic-AmbianceGTK.svg
+++ b/Numix/scalable/places/user-trash-full-symbolic-AmbianceGTK.svg
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   height="16"
+   viewBox="0 0 16 16"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.48.5 r10040"
+   sodipodi:docname="user-trash-full-symbolic_neu2.svg">
+  <metadata
+     id="metadata20">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs18" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1215"
+     inkscape:window-height="776"
+     id="namedview16"
+     showgrid="true"
+     inkscape:zoom="34.6875"
+     inkscape:cx="8.952187"
+     inkscape:cy="8"
+     inkscape:window-x="65"
+     inkscape:window-y="24"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg2">
+    <inkscape:grid
+       type="xygrid"
+       id="grid2997" />
+  </sodipodi:namedview>
+  <path
+     style="fill:#bebebe;fill-opacity:1;stroke:none"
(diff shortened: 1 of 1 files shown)


Foggalong:
Q: why is it just a red line and not a full red?

dirtydancing:
A full red is not displayed with Ambiance GTK theme at all, or more precisely displayed wrongly as a light grey, which is (almost) invisible on the light grey background. At first, I thought that with Ambiance GTK theme, no colours would be displayed at all with the symbolic icons in Nautilus/Files, because iirc when I changed the red to another red, it still was not displayed as red, but still as light grey. Then, I tested the red line variant suggested above, and this has worked.

To avoid any misunderstandings: this icon is not supposed to replace the current default full red Numix icon. It is only supposed to temporarily be of help for Numix icon theme + (Ubuntu default) Ambiance GTK theme users (this is the only "incompatibility" between Numix base and Ambiance GTK theme that I am aware of), until this bug is fixed from Ubuntu's side, cf. https://bugs.launchpad.net/ubuntu/+source/ubuntu-themes/+bug/1426774

wa4557:
@dirtydancing Can you make a red one with just a tiny spot grey, will that work?

dirtydancing:
Sure thing, I will look into it.

wa4557:
@dirtydancing It's probably even possible to make the grey spot outside of the canvas, this would be a cool hack for this bug :)

dirtydancing:
Grey spot outside of the canvas? Or some "hidden" residue? No idea how I should pull this off, but if such devilry is possible, this might even be a possibility for a revised default Numix icon, no need for an extra icon then.

@wa4557 Is it possible for you to test this on your machine? I am afraid this exceeds my current Inkscape skills :-)

wa4557:
@dirtydancing Sure, what do I have to do? Ambiance theme with Numix-base icon theme if I understand correctly, and the full-tray icon in scalable/status is the one that needs to be altered, correct?

cldx:
@dirtydancing you just make a pixel-sized box and drag it outside of the image border in inkscape. Then save.

dirtydancing:
@wa4557 Yes, Ambiance GTK theme with Numix-base, the icon is `user-trash-full-symbolic.svg` in `scalable/places` And once again, I am curious about what you will come up with :-)

@cldx Thanks for your help, appreciated. I am getting there. But right now, because there is so much going on with Numix Base, there a lot more issues that I can usefully comment on, so I will leave this one up to wa4557.

wa4557:
@dirtydancing I have a hard time figuring out what is going on here. Adding more red to your icon and still only the one line already added to your icon is shown. Make it completely red -> the icon disappears. Make it red except the line that is red now -> icon disappears.

Strange

dirtydancing:
Haha, welcome to my state of perplexity about this. The red line thing was more of a coincidence than anything else, but it turned out to work. This is why I believe some sort of Inkscape sorcery is needed here :-)

wa4557:
I think I figured out what's going on here: Only the contours of the different parts are shown in their color. The filling is not. What I'm doing with this information idk

Foggalong:
Wait... so it's only the outline parts that work, so to fix this we'd have to make an icon entirely out of outlines? I'm thinking this might be a 'wontfix', but someone really should report this problem to the Ambiance GTK team because if this is the problem it really is quite shoddy :/

wa4557:
Yeah exactly. I agree with the wontfix, this is just not doable (I actually tried it a little bit) + this is not a Numix issue imo

dirtydancing:
I am not sure what this "outline" thing means for Inkscape, or if this is the root of this problem :-) But I am happy to point to this conversation in the bug report against Ambiance GTK over at https://bugs.launchpad.net/ubuntu/+source/ubuntu-themes/+bug/1426774. Should I do that?

And as to the wontfix: of course, this is not supposed to be fixed from Numix's side at all, this is only an additional icon that works with Ambiance GTK, to be used by renaming at the discretion of Numix icon theme users who are also users of Ambiance GTK theme and who would like to see any trash-full icon in Nautilus/Files at all :-)

Foggalong:
When you draw a shape in Inkscape like this rectangle
screenshot - 290315 - 18 20 03
By default it's colour is determined by it's fill properties
screenshot - 290315 - 18 20 14
However it's also an option to add an outline to the shape, determined by it's stroke properties
screenshot - 290315 - 18 20 38
Which you can see changes to original shape accordingly
screenshot - 290315 - 18 20 47
Lines on the other hand like this one
screenshot - 290315 - 18 21 21
Are created with just a stoke property by default
screenshot - 290315 - 18 21 31
And no fill
screenshot - 290315 - 18 21 41

As I understand it, Ambiance is only displaying the red colour if it's part of the stroke rather than both stoke and fill too. As you can imagine it would be incredibly difficult to create a complicated shape like a trashcan just using the strokes.

So yeah, report the problem to Ambiance and hopefully they can do something to fix it :)

PR #310, opened 2015-03-21
Fixes inconsistent shadow for nm-signal-*

wa4557:
The shadows in Numix-Light and Numix both in 22x22/status and 24x24/status were inconsistent/incorrect. There was no distinction in the shadow between signal and non-signal bars, but the whole shadow was black with an opacity of 12% (non-signal shadow) (see screenshot)

Changed that accordingly to the right values (30% black for signal bars 12% for non-signal bars) for the Light and "Dark" theme

Before:
before

After:
after


wa4557:
Closing because there are no more shadows

PR #311, opened 2015-03-21
Bluetooth status icons

wa4557:
Fixes probably the last inconsistency I could find...

The bluetooth status icons (especially bluetooth-paired.svg) were inconsistent in 22x22/status -> fixed that. Moreover the icons in 24x24/status and 16x16/status were "strechted" because they were not scaled accordingly from 22x22/status -> also fixed 

plus support for Light theme


wa4557:
Closing, but will open a new pull-request since the inconsistency is still here

PR #313, opened 2015-03-21
Weather status icons

wa4557:
The weather status icons right now are the ones from ubuntu-mono-dark, so I thought that it would be nice to have proper numix weather status icons (the app "indicator-weather" relies on them for example). I used the icons from Numix/scalable/status aligned them, colored them accordingly and placed them in 22x22/status 18x18/status and 24x24/status (for light and dark). 

Seems to be a big pull request; but there are actually just a few files and a lot of symlinks

If you don't like them I'm not mad if you don't merge my pull-request.


wa4557:
Closing because new status icons. Will eventually create new pull request, because the weather icons are still the ones from ubuntu-mono-dark

Issue #286, opened 2015-03-10
Warning red colour inconsistency with the status icons


dirtydancing:
When looking into a couple of "warning red" status icons, I came across another colour inconsistency after the "light grey" colour inconsistency reported at https://github.com/numixproject/numix-icon-theme/issues/282

The "warning red" colour appears to be either <a href="http://www.color-hex.com/color/ef555c">#ef555c</a> or <a href="http://www.color-hex.com/color/ed535b">#ed535b</a>

Two examples for #ef555c

![syspeek-100](https://cloud.githubusercontent.com/assets/7164227/6574509/0a88d12c-c726-11e4-92d6-51401fa19323.png)

![battery-empty-symbolic](https://cloud.githubusercontent.com/assets/7164227/6574511/121c9cac-c726-11e4-9879-5b87f2dc13fa.png)

Two examples for #ed535b

![error](https://cloud.githubusercontent.com/assets/7164227/6574515/1f0f07e2-c726-11e4-8595-eab038fd7af1.png)

![xfce-system-lock](https://cloud.githubusercontent.com/assets/7164227/6574518/22ea7130-c726-11e4-845e-cbbdc9141806.png)

Maybe it would be a good idea to create a Numix (base) colour overview for quick reference and for consistency.

dirtydancing:
@Foggalong Coming back to the idea above to create a Numix (base) colour overview: Taking the example from https://github.com/numixproject/numix-folders, I have been experimenting with that for the readme file of numix-icon-theme, cf. https://github.com/dirtydancing/numix-icon-theme/tree/colour-chart.

Screenshot:

![numix-icon-theme_readme-colours](https://cloud.githubusercontent.com/assets/7164227/6765262/8e3fcb78-cfd8-11e4-9d12-bf7d26de1914.png)

Does this make any sense at all, and should I create a pull request for this?

wa4557:
Cool! 
But you're missing the "inactive" (the color of your audio status icon, if you mute your audio completely for example) color somehow. 

40%opacity #dcdcdc for light grey
and 
40% opacity #353535 for dark grey 

plus you could also add the color of the shadow (if there's one); 30% opacity black for all the colors in your post and 12% opacity black for all the "inactive" colors...

dirtydancing:
[...]

Issue #282, opened 2015-03-06
Colour inconsistency with the panel status icons


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme-circle/issues/1380#issuecomment-76214861 as well as in relation with https://github.com/numixproject/numix-icon-theme/issues/271 I came across a slight grey colour inconsistency with the panel status icons.

I figure this colour should be consistent, already to avoid unnecessary duplicate work e.g. with a future set of indicator-keyboard icons.

In general the colour here appears to be <a href="http://www.color-hex.com/color/dcdcdc">#dcdcdc</a> (gainsboro), but for some panel status icons, e.g. for indicator-messages and indicator-keyboard, it is <a href="http://www.color-hex.com/color/dedede">#dedede</a> (gray87).

Possibly the #dedede is the new Numix panel icon colour? These two colours almost look the same, but if I had to choose I would probably go for #dedede, no idea why. Ok, I will stop speculating here :-)

Screenshots in Inkscape with icons from `Numix/22x22/status` (system status icons) and from `Numix-Circle/22x22/status` (app status icons):

indicator-messages #dedede

![222_indicator-messages](https://cloud.githubusercontent.com/assets/7164227/6524352/cb46b1d0-c3f8-11e4-8195-bf1ce5042eba.png)

indicator-keyboard-En #dedede

![222_indicator-keyboard-en](https://cloud.githubusercontent.com/assets/7164227/6524354/d3257ec2-c3f8-11e4-95f5-c8ee63f3fdae.png)

shutter-panel #dedede

![222_shutter-panel](https://cloud.githubusercontent.com/assets/7164227/6524356/d84a5c56-c3f8-11e4-8e47-35aad03d4f00.png)

And in contrast, examples of status icons with #dcdcdc

system-devices-panel

![220_system-devices-panel](https://cloud.githubusercontent.com/assets/7164227/6524360/dd897300-c3f8-11e4-8127-8a179af8628c.png)

audio-volume-high

![220_audio-volume-high](https://cloud.githubusercontent.com/assets/7164227/6524363/e25a5eb2-c3f8-11e4-9ccd-ae1640576b9c.png)

battery-100

![220_battery-100](https://cloud.githubusercontent.com/assets/7164227/6524366/e88dfbea-c3f8-11e4-83db-9b630d849ca4.png)

caffeine-cup-full

[...]

PR #321, opened 2015-03-24
Added Colour Chart

dirtydancing:
Following-up on https://github.com/numixproject/numix-icon-theme/issues/286 and https://github.com/numixproject/numix-icon-theme/issues/282, this is a colour chart placed at the end of the Readme.

Screenshot:

colour-chart


wa4557:
@dirtydancing I think this needs some changes, doesn't it :)?

dirtydancing:
Yeah, will do :-)

wa4557:
Btw: You're missing the blue color from the `indicator-messages-new.svg`

dirtydancing:
Good occasion to include that as well then, thanks.

Foggalong:
This is a nice idea, but shouldn't be on the main readme that the users see. With the redesign anyway all the colours should be the same for status icons :)

dirtydancing:
Understood. This is indeed not pointing to the users. Actually, I asked a user for some feedback on this, and he thought this might be a sort of "reset button" (?) to get back to the defaut colours. No idea if that interpretation is representative, but this confirms that the main readme is not the place to put this on. Right now, I cannot think of another place to put this on, either. And with the redesign, there is consistency and more simplicity (no shadow) anyways.

So I am closing this.

Issue #107, opened 2014-04-19
Big Battery Status Icon


angelolev:
I am using elementary os luna and when i installed this theme, i got a huge battery icon, how can i fix it? thanks

![icon](http://imgur.com/2O5FPk6)

fselem:
@dirtydancing I recently installed the second beta for elementary os Freya, and the issue seems to be completely resolved.

![screenshot from 2015-02-12 03 26 18](https://cloud.githubusercontent.com/assets/7390659/6164007/9b63743c-b267-11e4-9374-453e3c9ca12b.png)

![screenshot from 2015-02-12 03 27 09](https://cloud.githubusercontent.com/assets/7390659/6164008/a19910dc-b267-11e4-972a-f1b1e64868f8.png)

Also, check out the new Beta 2 I was talking about http://beta.elementaryos.org

dirtydancing:
@fselem Great! Thanks for your feedback. I shortly did check Freya Beta 2 after it was released, and I did not experience any battery icon size issues either.

Not relating to the battery icon size, just to let you know: One other thing I noticed is that the colour of the "fully-charged" battery icon is somewhat lighter with the default GTK theme (cf. your first screenshot), but e.g. with Numix GTK theme corresponds to the colour of the other indicator icons.

dirtydancing:
@fselem PS: I just checked the battery icon colour again, and it is whiter not only for fully-charged, but also for charging (not for discharging, though, as seen in your second screenshot above). Again, with the default elementary GTK theme, not e.g. with Numix GTK theme. This appears to be related to the primary use of "symbolic"-named battery status icons for fully-charged and for charging (but not for discharging) in Ubuntu 14.04 LTS, which Freya is based on, and this seems to be a GTK theme issue.

No issues with the battery status icon size.

Comparative screenshots (battery charging):

Default elementary GTK theme:
![elementary_gtk](https://cloud.githubusercontent.com/assets/7164227/6164932/0038fbc0-b2a4-11e4-9834-78ff286c928d.png)

Numix GTK theme:
![numix_gtk](https://cloud.githubusercontent.com/assets/7164227/6164935/084cfd2a-b2a4-11e4-893e-8e8da94ac3f7.png)

dirtydancing:
[...]

PR #333, opened 2015-03-26
Fixes #331

### LINKED ISSUE #331: Additional percentage level battery status icons in fixed size directories 16, 22 and 24
Following-up on my comment over at https://github.com/numixproject/numix-icon-theme/issues/107#issuecomment-76519738, I have done some testing with the additional percentage battery status icons introduced for Ubuntu 15.04 for 10%, 30%, 50%, 70%, 90%:

These (tested with testing battery icons size 22x22) are already working in current daily Ubuntu 15.04, and at the same time they do not cause any breakage in Ubuntu 14.10, which simply does not pick up these additional battery status icons. Also tested against elementary OS Freya Beta 2, based on Ubuntu 14.04 LTS: just like in Ubuntu 14.10, the additional battery status icons simply do not get picked up.

These additional icons are useful because they provide a finer granulation, and because they address the issue of a too early warning status with the present icons, cf. bug report at https://bugs.launchpad.net/ubuntu-ux/+bug/1388235.

Screenshots when tested simultaneously against Ubuntu 14.10 (above) and Ubuntu 15.04 (VM, below):

battery_28

battery_20

battery_14

battery_9

battery_4

battery_charging-2

battery_charging-5

battery_charging-21

battery_charging-35

battery_charging-90

This is how these additional battery status icons (10%, 30%, 50%, 70%, 90%) could be fit into the present ones, which in this overview have been left untouched (additionally all selected for visual contrast).

battery-new-levels

battery-new-levels2

I would create a pull request for these myself, but I am not yet very familiar with Inkscape (for the icon overview above I just scaled the already existing bars, which of course does not work for the charging status battery icons).

@wa4557 Are you interested in contributing these? For bar height in the screenshots above, I used px (new sizes bold) **2,5** orange - 4 orange - **5,5** - 7 - **8,5** - 10 - **11,5** - 13 - **14**. In addition, the battery icons seem to use #dedede instead of #dcdcdc, so this might be an opportunity to make these consistent as well. From my perspective, adding these battery status icons makes sense even if there is a status icons redesign in the future, because this establishes a baseline on which such redesign could build upon.

The additional battery status icons in question here are marked yellow in this overview:

ca8cdf3c-bf37-11e4-8d9b-a2d16276d471

wa4557:
Adds more percentages for the battery icons cf. #331 
Already new design, for 16x16 22x22 24x24 for light and dark theme

### DIFF
--- a/Numix-Light/16x16/status/battery-010-charging.svg
+++ b/Numix-Light/16x16/status/battery-010-charging.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   viewBox="0 0 16 16"
+   height="16"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.91 r"
+   sodipodi:docname="battery-010-charging.svg">
+  <metadata
+     id="metadata14">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs12" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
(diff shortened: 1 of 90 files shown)


dirtydancing:
Wow, faster than fast, and already in the new design :-) Great!

Two points:

1) I noticed that there are 58 files changed, but it should be 60. Any reason why in `Numix-Light/16x16/status`, `battery-010.svg` and `battery-090-charging.svg` are missing?

2) Bonus request: could you add the gpm-battery symlinks (`gpm-battery-...` symlink with the target `battery-...`) for each of these new icons, as listed in the overview below? The symlinks further down are not needed for these additional icons at the moment, as far as I can tell. Let's go with the gpm-battery symlinks for now.

This only applies to Numix, not to Numix-Light (consistency also here, in Numix-Light all of these symlinks do not exist as of now anyways, iirc).

symlinks_gpm-battery

wa4557:
1) No just missed them, added
2) done

dirtydancing:
Thanks!

dirtydancing:
Updated overview over at https://github.com/numixproject/numix-icon-theme/issues/107#issuecomment-87110450

Foggalong:
I have to leave that page to load all the images for a few minutes now because otherwise it just jumps around making it unreadable :P

dirtydancing:
Yeah, I can confirm the loading time for that page :-)

PR #339, opened 2015-03-27
Aligns owncloud icons

wa4557:
The owncloud status icons (mostly status-offline.svg) were not properly aligned and had the wrong color. Moreover I changed one regular file to a symlink since it should be the same


Foggalong:
ownCloud client and "state-ok" are used in different contexts so need keeping as distinct icons

wa4557:
Ah ok, I think that was not on purpose I'll change that

wa4557:
Reverted, should be ok now

PR #347, opened 2015-03-27
Fixes wrong scaling of nm-vpn-standalone-lock.svg

wa4557:
Same as #342 I think (and hope) that's the last one


Foggalong:
You seem to have switched this one to incorred scaling :/ it's supposed to be 16px

wa4557:
Erm no? The original icon (the one in 22x22/status) is 22px times 20px. Therefore the icon in 16x16 should be 22px times 16/22*20px, and not 16px*16px, am I wrong?

Foggalong:
Ah, I see! Got my height and width mixed up :P

PR #348, opened 2015-03-27
Adds notification icons

wa4557:
New pull request, everything was already said in #298 
One thing though: I used #ececec instead of #dcdcdc to be consistent with the new status icons


dirtydancing:
Already merged, great! When I was recently doing some testing with Numix base, I found the non-Numix notification icons quite irritating.

PR #356, opened 2015-03-28
Fixes inconsistent transmission tray icon design

wa4557:
The design for the transmission tray icon was different for different sizes (16 and 24 looked different to 22). Moreover the 24x24/status icon was in 24x24/apps instead of status

Edit: just recognized, that the design of the icons is also different; before merging I should fix that as well.
@Foggalong which one is the one you want to keep: the one in 22 and 16 or the one in 24?


Foggalong:
@wa4557 This is the desgin we want to keep :)

wa4557:
Ok, consistent design in all three sizes now

PR #365, opened 2015-03-28
Fixes #362

### LINKED ISSUE #362: Request: Mimetype: application/x-subrip
Name: **SubRip subtitles**
Mimetype: **application/x-subrip**
Possible extensions: **.srt**

Description: **File responsible for provide subtitles for movies/videos** (very common for non-English speakers)
image
Target File name: application-x-subrip.svg

hlechner:
#362 
For some reason the github preview doesn't show the properly icon.

@palob Do you know if I need to modify something else?

### DIFF
--- a/Numix/128x128/mimetypes/application-x-subrip.svg
+++ b/Numix/128x128/mimetypes/application-x-subrip.svg
@@ -67,31 +67,31 @@
      inkscape:window-width="1920"
      inkscape:window-height="1016"
      id="namedview10"
-     showgrid="false"
-     inkscape:zoom="3.4766083"
-     inkscape:cx="-36.959623"
-     inkscape:cy="23.731712"
+     showgrid="true"
+     inkscape:zoom="2.8284271"
+     inkscape:cx="-63.675791"
+     inkscape:cy="34.524776"
      inkscape:window-x="0"
      inkscape:window-y="27"
      inkscape:window-maximized="1"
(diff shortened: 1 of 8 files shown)


Foggalong:
Make sure you put any straight parts and align them to grid :)

hlechner:
Actually I've no idea how to do that :P

I thought that I've aligned it before. So I need to learn it, probably I'll spend some time before finish it

Foggalong:
Press `#` to make sure the grid is showing, then press `F2` and click on the shape you want to align. You'll see little dots called nodes appear. Just make sure that the outline of the shape (that isn't a corner) between two nodes is following the lines of the grid :)

hlechner:
Could you let me know if I made it right in the 48x48? so I can apply to the rest

palob:
For the preview showing broken icons, I've had this before when editing or using elements from existing Numix Circle icons (presumably the ones which already existed before the base plate conversion). I haven't understand what's casuing this so far.

Foggalong:
@hlechner yup, that's aligned right :)

hlechner:
Done, but should be an easier way to do that than edit one by one :P

palob:
Yep, everything's aligned+centered now :+1:

Foggalong:
That's the downside of doing icons for every size, and why we only do 48px for Numix Circle :)

PR #376, opened 2015-03-28
deletes connection animation in 22x22/apps

wa4557:
Deletes the connection-animation from 22x22/apps since it is already present in 22x22/animations, has nothing to do in an apps directory and is wrong anyway... Is there a reason why you kept it there? I think that was an error...


Foggalong:
It was done twice, first authored [here](https://github.com/numixproject/numix-icon-theme/commit/559556571d876411525317227d5b67790134d55b) last February to `/apps` and then again [here](https://github.com/numixproject/numix-icon-theme/commit/8b23c0ffa6f080c64e02e271eec7c9a9b7b16add) in June. I have no idea why, but I'm guessing it's fine just to have them in animations.

Foggalong:
Just done a quick check and even though the icon names are the same the animations are different :/ I have no idea which it is that's currently displayed in Numix

screenshot - 280315 - 22 44 27

wa4557:
Well at least in Ubuntu the one in `/animations` is enough (don't know if other DEs use these icons too). Morever the design of the left one is incorrect, so if they are needed we should symlink them

dirtydancing:
Are you sure that Ubuntu 14.10 uses the ones in `22x22/animations`? I just checked, and for me Ubuntu 14.10 seems to take those from `22x22/apps`. They are a little inconsistent (too big), but they are displayed. When deleting the ones in `animations`, I encountered no difficulties. In contrast, when deleting the ones in `apps`, no animation was displayed (first shortly a question mark, then no icon at all, then the status icon). This was just a quick check.

Anyways, this shows to me that it is probably a good idea to keep these icons in both `animations` and `apps`, to adjust the design and to use symlinks. Please let me know if I should test more profoundly.

wa4557:
@dirtydancing Just tested it, and you're right. Symlinking seems to be a good idea, just to be safe. I wonder if it's necessary to add animations in the 24x24 and 16x16 status directories as well (probably yes)

dirtydancing:
About the sizes 24x24 and 16x16: I have not checked, but for reasons of consistency I would say yes, this is probably a good idea to add them as well. Just to avoid any misunderstandings, please correct me if I am wrong: those would not be in the `status` directory, but rather consistently in the `apps` and `animations` directories.

wa4557:
Yeah in the apps and animations directories. Ohne thing though: isn't the animation in Ubuntu-mono-dark also placed  in animation directory (I can't check right now)? If yes why are the icons taken from the apps directory?

dirtydancing:
Yes, Ubuntu-mono-dark has these icons in `animations/22` and in `animations/24` (although it seems that the animation icons in `22` also have the canvas size 24x24, well, not of relevance here). No idea why these are taken from the `apps` directory in Numix base. All listed in `index.theme` in Numix base, so this is not a possible explanation. Strange.

Foggalong:
The more time I spend working on icon themes the more I start to think that there's no logic at all to the way any of this works and that everything is simply held together with super glue and duct tape.

dirtydancing:
@Foggalong Haha, super comment! Might actually apply not only to icon themes :-)

wa4557:
At least we're not running out of bugs to fix :) Ok I'll do that as soon as I'm happy with the status icons

wa4557:
@dirtydancing Ok I tested profoundly, and I can NOT confirm that deleting the animation in 22x22/apps leads to not displaying a sequence at all. You're right that the icons are first taken from 22x22/apps (animation is too big and ugly), but if I delete them the correct animation (the one from the 22x22/animations) directory is shown. 

If you can confirm this I suggest merging this pull request. Icon.cache once more maybe?

dirtydancing:
Yep, I did some more profound testing, and I can confirm this. No icon.cache though :-) So yes, this works, and the icons in `animations` are much better (and those in `apps` not needed). And I agree that more generally, such animations icons make no sense in the `apps` directory.

In conclusion, I also recommend merging this pull request.

wa4557:
Cool!

dirtydancing:
btw, animations in Numix-**Light** also already work great with the latest Numix base. Well done!

wa4557:
I close this and make a pull request with deleted in `apps` and animations in 24x24 and 16x16

PR #390, opened 2015-03-29
Adds symlinks in Numix-Light

wa4557:
This adds all the symlinks that are present in numix-base to numix-light (and the icon they are pointing to is present of course). This seems to be a huge pull request, but no single icon was changed or added in this pull request, only symlinks added. 

Edit: actually there are three files, that gets deleted with this pull request, but they are replaced by the proper symlink

@Foggalong If you merge this (and my previous pull requests) I think the status icons are as good as they can get (for the ones there right now). All icons (except the battery ones which I don't want to touch :) are present in all three sizes 16 22 and 24, all color and scaling inconsistencies are ironed out and numix-base and numix-light are basically the same.


Foggalong:
Excellent! :D Thanks for not touching the batteries too, because the indicators for various distros are programmed so poorly it's a delicate balancing act getting them all working :P

PR #409, opened 2015-03-30
Vala mimetype enhacement

palob:
Existing one was the hasbang icon used for shell scripts etc.
Now with the vala symbol from Numix Circle.
text-x-vala2


Foggalong:
Good idea!

Issue #258, opened 2014-12-15
Synapse indicator icon


bil-elmoussaoui:
Hello,
a synapse indicator icon is missed, i found an icon on /usr/share/icons/Numix/16x16/status/synapse.svg
but its not looking the same on my appindicator 
![selection_001](https://cloud.githubusercontent.com/assets/7660997/5435164/11cabd26-8458-11e4-8982-5fa90fb48c75.png)
i really dont think its hardcoded ;
![menu_003](https://cloud.githubusercontent.com/assets/7660997/5435190/62491f7c-8458-11e4-84e2-f61398a04f78.png)
this picture is from synapse indicator too and it's using the themed icon... maybe you should add an other size? like 24x24? 
thanks

wa4557:
I don't think this is fixable, since it's an icon from scalable. So it has automatically a different design and color. This is a synapse bug imo...

bil-elmoussaoui:
I don't think so, i'm using elementary OS now with the elementary base theme and the icon is picked from actions/24x24

wa4557:
@bil-elmoussaoui Yeah but it doesn't make sense to add an *symbolic icon to a non scalable folder. So if synapse is using `edit-find-symbolic.svg` that is and should be an icon from `scalable` and consequentely has a different design

bil-elmoussaoui:
Well, we might add the same icon from scalable into the 24x24 size and give it a try? I've read somewhere in the elementary community (Google+) that this icon must be a 16x16 icon in a 22x22 shape or something like that; because the icon looks big in elementary too.
 If someone want to report this to the developer of synaptic indicator it will be very helpful.

PR #410, opened 2015-03-30
Fixes #405

### LINKED ISSUE #405: Wrong Transmisson style
@wa4557 when I told you which style to update the transmission icons to you did the wrong one :P This is the style that's wanted, the 16px version was just like it was because the arrow didn't scale very well. Sorry for the confusion!
screenshot - 300315 - 12 13 10

wa4557:
Fixes wrong design of transmission icons. Doesn't scale too bad imo.
+ One icon that was wrongly scaled in 24x24/status (synapse.svg)


bil-elmoussaoui:
I believe the synapse indicator dont use the synapse icon. You can look at this for more information #258, so for me the synapse icon should be removed from status icons ;)

Foggalong:
@bil-elmoussaoui the synapse indicator will be there for a reason so it's just safer to leave it in

PR #412, opened 2015-03-30
Added mimetype for the R language

palob:
text-x-r

### DIFF
--- a/Numix/128x128/mimetypes/application-x-r-project.svg
+++ b/Numix/128x128/mimetypes/application-x-r-project.svg
@@ -1,0 +1,1 @@
+text-x-r.svg
--- a/Numix/128x128/mimetypes/text-x-R.svg
+++ b/Numix/128x128/mimetypes/text-x-R.svg
@@ -1,0 +1,1 @@
+text-x-r.svg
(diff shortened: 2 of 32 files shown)


Foggalong:
Would it be possible to incorporate the design of RLogo.svg from Circle?

palob:
It is the *R* from Circle. So you're talking about the lines? Might be doable for 32px+.

Foggalong:
Yeah, the swirly lines :)

palob:
A few quick drafts:
text-x-r-lines
text-x-r-lines3
text-x-r-lines5
text-x-r-lines32
text-x-r-lines64

Didn't turn out quite as well as I thought it could. Looks a bit busy, doesn't it?

Ghokun:
How would it look like without the R letter, just with the ellipsis.

Foggalong:
It would look unrecognisable as an R icon :P That first draft, how would it look if the symbol was right 1px and down ~2px?

palob:
text-x-r-lines0

Foggalong:
I really like that :D

palob:
OK, let's see how it projects to other sizes.

Foggalong:
Sure thing :)

palob:
plain
swirl

Plain *R* for 16, *R+ellipsis* for bigger sizes?

Foggalong:
I think you can probably get away with the swirl for all sizes. A niche enough icon that those needing it will recognise it

palob:
Swirls are up :)

PR #415, opened 2015-03-31
Added mimetype for Octave files

palob:
Pushed design, same colour as formula mimetype:
octave
Alternative colour (from cubes in Octave logo):
octave2


Foggalong:
Alt colour :D

palob:
Up :+1:

PR #448, opened 2015-04-11
Adds workspace indicator status icons fixes #296

### LINKED ISSUE #296: Workspace indicator
Hello, 
i would like to request some icons as a support of workspace indicator 
the icons from the workspace 1 to 36, you can make a python script to generate other files (i can do that) 
The icons are on this folder : /usr/share/icons/hicolor/16x16/status
the icon names are : indicator-workspaces-X  with X = 1, 2, ..., 36
the icon looks like that
selection_002

Thanks a lot!!!

wa4557:
#296


Foggalong:
Again needs aligning to grid, definitely the monitors and as much of the numbers as is possible without making them look silly (characters are always a pain so don't worry to much about these)

wa4557:
Ok aligned. Inkscape verbs are quite powerful!

Foggalong:
Good work on 22/24, but 16 still needs doing. If you feel the outline needs to be more than 1px, then at least one of the two (probably outside here) edges needs aligning to grid.
screenshot_2015-05-02_14-10-52

wa4557:
Yeah I made an error the last time I pushed I missdid 16 somehow. Should be ok now

Foggalong:
I'm looking at it and I think you can probably just go full on alignment with a 1px border.
screenshot_2015-05-02_15-20-20
The change in clarity seems to be worth it 
screenshot_2015-05-02_15-20-29

wa4557:
Fine but I will do that tomorrow, that's quite some work because it can't be automated :/

Foggalong:
Ah, fair enough :) Great job though on these, nice idea having them on the monitor too

wa4557:
Ah cleverly using sed it was less work than expected :)

Foggalong:
Thanks for the great work on this one <3

PR #451, opened 2015-04-11
Adds battery icons in 32x32/devices fixes #206

### LINKED ISSUE #206: Battery icon: remove battery-good-charging.svg from devices directories
Hi, question: is there a reason why currently there is a symlink/icon (depending on size) named `battery-good-charging.svg` in all of the fixed size `devices` directories? There is an icon named `battery.svg` in devices which belongs there.

Screenshot from `48x48/devices`:

batter-good-charging

`battery-good-charging.svg` is not a device icon, but, if at all (that would be a different issue) a status icon. Moreover, the design is that of a "neutral" battery and does not indicate a charging status (flash symbol, battery filled to a certain height).

So, unless there is a specific reason for this icon in `devices`, I would suggest to remove it.

Additional information: In Ubuntu 14.04 LTS, these icons are used e.g. for the "Energy Statistics" window which is not in the panel, but can be accessed when clicking on the battery symbol in the panel power dropdown window. Here are two screenshots, the first one showing the (incorrect) Numix `battery-good-charging.svg`, the second one showing the GNOME fallback icon for `battery-full-charging.svg`.

energy_1

energy_2

This observation actually goes back to an observation in this former comment https://github.com/numixproject/numix-icon-theme/issues/107#issuecomment-52829244

wa4557:
Adds battery icons in 32x32/devices and ac-adapter.svg in 32x32/devices for discussion see #206


wa4557:
Argh completely wrong, sorry

Issue #236, opened 2014-10-31
Battery icon 10: in fixed size directories 16, 22 and 24, delete the 2 symlinks "battery-caution.svg" and "battery-low.svg"


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/issues/226 and https://github.com/numixproject/numix-icon-theme/issues/107, step 10:

In fixed size directories 16, 22 and 24, delete the 2 symlinks `battery-caution.svg` and `battery-low.svg`.

These 2 symlinks are not needed for the panel, and in this context they only cause confusion.

![battery-status-icons_20141028_carried-over_10](https://cloud.githubusercontent.com/assets/7164227/4858786/fcdf1a76-60e2-11e4-806a-caf9943920cb.png)

Now, these are not outdated icon names. They are just used for a different purpose, at least in Ubuntu 14.04 LTS/Unity: They are NOT used for the panel (monochrome), but rather e.g. for the gnome-power-statistics window of the gnome-power-manager (Ubuntu). Ubuntu default (just as Numix base) currently is using the GNOME fallback icon theme here.

Screenshot for "full-charging" status, with current Numix-Circle, please note that although this is a device information list, for Ac adapter, a static device icon is used, whereas for Laptop battery, dynamic status icons are used:

![statistics_full-charging](https://cloud.githubusercontent.com/assets/7164227/4858889/2a8f72b2-60e4-11e4-8924-4f5090462091.png)

With this, it becomes apparent that the current symlinks `battery-caution.svg` and `battery-low.svg` actually are visually unfitting for this purpose. Screenshot with current Numix-Circle for "battery-low" status (icon is hardly visible):

![statistics_low](https://cloud.githubusercontent.com/assets/7164227/4858901/48781252-60e4-11e4-8a59-abc527df4788.png)

If this series of battery icons (NOT "percentage"-named, NOT "symbolic"-named) is used, then there would have to be a consistent and encompassing set (empty to full-charged), not only the "stand-alone" icons `battery-caution.svg` and `battery-low.svg`. Again, that would be a different issue, so I am leaving it at that for this issue.

Side note: At the same time, this also shows that https://github.com/numixproject/numix-icon-theme/issues/206 is still valid and that `battery-good-charging.svg` should be removed from the `devices` directories (not a panel issue, though).

This step 10 concludes the mini-series of step-by-step issues. With steps 1-10, present Numix base is modified to arrive at the current battery status icons diagram:

[...]

PR #452, opened 2015-04-11
batteries in 32x32/devices / ac-adpater fixes #206

### LINKED ISSUE #206: Battery icon: remove battery-good-charging.svg from devices directories

wa4557:
Adds battery icons for the ubuntu power management menu + ac-adapter.svg for discussion cf. #206


dirtydancing:
How about removing the current `battery-good-charging.svg` not only from `32x32/devices`, but rather from all `devices` directories? It is the wrong icon for that name in all sizes. Should that be dealt with in a separate pull request?

And would a separate pull request also be the way to go to remove `battery-caution.svg` and `battery-low.svg` from the `status` directories in fixed sizes 16x16, 22x22 and 24x24, because these icon names belong to the "battery device status" set of battery icons? Corresponding issue report: https://github.com/numixproject/numix-icon-theme/issues/236.

wa4557:
Yeah probably. Fancy doing that? I can do it as well if you want too :) One additional thing: I should probably add the above icons also to other sizes (48x48 at least, ac-adapter to all other sizes?, adding them to 16x16 22x22 and 24x24 is probably a bad idea)

dirtydancing:
Yes, please go ahead with all these corresponding pull requests, this is the easier way to handle this.

And yes, `ac-adapter.svg` makes sense for other fixed sizes (not the `scalable` directory though). In gnome icon theme shipping with Ubuntu, `ac-adapter.svg` is in the fixed size `devices` directories up to 48x48. That might be a good reference point for Numix base as well (thus 16, 22, 24, 32, and 48). Currently I am not aware of a conflict that could come up in 16, 22 or 24, as those contain `ac-adapter-symbolic.svg` (not in 32 and 48, though).

Alternatively, just adding `ac-adapter.svg` to `32x32/devices` and to `48x48/devices` would make sense, too, as apparently this icon has not been needed in 16, 22 or 24 so far :), and there is less chance of a (potential) conflict.

In Numix base, starting with 64x64, the `devices` icons appear to change anyways, turning rather monochrome (?).

About the sizes for the "battery device status icons": agreed that it makes the most sense at the moment to have them not only in `32x32/devices`, but also in `48x48/devices`, (but not in 16x16, 22x22 or 24x24).

Giving this another thought, in conclusion and for the time being, I would recommend to add `ac-adapter.svg` and the set of "battery device status icons" to 32x32 and in addition to 48x48.

wa4557:
Fine, everything is in place. From my point of view this may be merged :)

dirtydancing:
@wa4557 Is there a reason why the current `battery-good-charging.svg` is still present in sizes 64x64, 128x128 and 256x256? I would suggest to remove the current `battery-good-charging.svg` (identical to `battery.svg`) from all sizes.

dirtydancing:
I am hopeful that this pull request is not overloaded, because it contains the battery "device status" icons as well as the added `ac-adapter.svg`. They cover related issues, so putting these together does make sense. To provide a visual impression, here are two comparative screenshots of "Energy Statistics", with current gnome fallback and with Numix as proposed here (Ubuntu 14.10, battery status full-charged):

current gnome fallback:
gnome

Numix as proposed here:
numix-proposed

wa4557:
Looks good :)

dirtydancing:
Agreed :)

Foggalong:
@dirtydancing @wa4557 why has `battery-good-charging.svg` been removed from 16, 128, and 256?

wa4557:
@Foggalong Because it's a `status` icon and belongs there, not a `device` icon

dirtydancing:
Agreed, and in addition this is also for consistency.

Foggalong:
[But then it's added in some sizes like 32px?](https://github.com/wa4557/numix-icon-theme/blob/bat32/Numix/32x32/devices/battery-good-charging.svg)

wa4557:
Yeah that's incorrect. They i'll remove them there as well

wa4557:
No this is not an error. It's just because all the other battery `devices` icons are present in 32x32 and 48x48 (because the battery icons are somehow special)
So consequentely battery icons after this pull request are 

`battery-caution-charging.svg` ` battery-empty.svg` `battery-full.svg` `battery-low-charging.svg` `battery.svg` `battery-caution.svg` `battery-full-charged.svg` `battery-good-charging.svg` `battery-low.svg` `battery-empty-charging.svg` `battery-full-charging.svg` `battery-good.svg` `battery-missing.svg` but only in 32 and 48 and nowhere else

Foggalong:
Why only in those too sizes?

wa4557:
Because the other sizes would interfere with the battery icons in smaller sizes (16 22 and 24). Yeah the battery icons are indeed a special case

dirtydancing:
These are the infamous battery `device status` icons that are not displayed in the panel, thus no sizes 16x16, 22x22 or 24x24 (could if at all only cause problems). They are used in size 32x32 for the Ubuntu "energy statistics", and it made sense to include them in size 48x48 as well. No larger sizes needed.

Foggalong:
@dirtydancing I thought you'd told me previously that these icons were never used as status icons in any size?

dirtydancing:
For clarification, please cf. this draft overview:
draft

dirtydancing:
@Foggalong Not sure when I said that, but I know one thing when it comes down to the battery icons: never say never :)

Foggalong:
That graph has just confused me more. Why are we removing the icons from 128px if we're adding them in in 32px and 48px?

dirtydancing:
Which icons are removed from  128px? Should only be `batter-good-charging.svg` (?).

wa4557:
Because previously the `battery-good-charging.svg` was just a symlink to `battery.svg` which makes absolutely no sense, because one should be a charging icon and the other one just a generic battery icon -> thus removing. 
Adding all these battery icons to bigger sizes makes no sense as well, because making them monochrome leads that ALL the icons will look the same (there wouldn't be a difference between `good` `low` `empty` and so forth). Plus they are not needed. Thereferore there is only the generic `battery.svg` in bigger sizes

dirtydancing:
And we are adding them because, while these battery icons are in the `device` category, they also show a status, cf. this comment https://github.com/numixproject/numix-icon-theme/issues/206#issuecomment-87053951

PR #453, opened 2015-04-12
Removes unnecessary battery symlinks from 16 22 and 24

### LINKED ISSUE #236: Battery icon 10: in fixed size directories 16, 22 and 24, delete the 2 symlinks "battery-caution.svg" and "battery-low.svg"

wa4557:
In continuation of #206 and #452 and @dirtydancing  asked me to do it, this fixes #236

### DIFF
--- a/Numix-Light/16x16/status/battery-caution.svg
+++ b/Numix-Light/16x16/status/battery-caution.svg
@@ -1,1 +1,0 @@
-battery-020.svg
--- a/Numix-Light/16x16/status/battery-low.svg
+++ b/Numix-Light/16x16/status/battery-low.svg
@@ -1,1 +1,0 @@
-gpm-primary-040.svg
--- a/Numix-Light/22x22/status/battery-caution.svg
+++ b/Numix-Light/22x22/status/battery-caution.svg
@@ -1,1 +1,0 @@
-battery-020.svg
--- a/Numix-Light/22x22/status/battery-low.svg
+++ b/Numix-Light/22x22/status/battery-low.svg
@@ -1,1 +1,0 @@
-gpm-primary-040.svg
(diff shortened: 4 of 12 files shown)


dirtydancing:
@wa4557 Thanks!

PR #454, opened 2015-04-12
Adds applications-fonts for all sizes fixes #237

### LINKED ISSUE #237: */categories/applications-fonts.svg empty
I think the title says it all: in any size the applications-fonts.svg is an empty svg.

Used to be a letter A?

wa4557:
Adds applications-fonts to all sizes. I took the design from two years ago from a commit that still had it :)
https://github.com/numixproject/numix-icon-theme/blob/eb559cecc6b759a45be2d21c89b402a0ee3b3a31/Numix/categories/48/applications-fonts.svg


Foggalong:
That's a very old mimetype icon from before we switched to the sheet design. This is our fonts icon which is already in the theme - are you sure you got the icon name right?

screenshot_2015-05-01_22-26-06

wa4557:
But this is no mimetype but a categories icon. Right now the icon is just an empty icon, which might be eaten by inkscape at some point

Foggalong:
Oooo. My bad :P

PR #456, opened 2015-04-12
Consistency between filenames for project(s) folder icons

figgyc:


wa4557:
Cool! but you should create this pull request in this repo: https://github.com/numixproject/numix-folders (for all 6 styles preferably). It will then be merged into this repo

dirtydancing:
May I ask: what is the problem here? I have been using the `projects` folder for quite a while, and iirc that has always been named `projects` with an "s" (which I think is preferable anyways).

figgyc:
@wa4557 Will do. @dirtydancing The name for every other resolution is project and for 16x16 it's project**s**.

dirtydancing:
Hmm, strange, I have seen `projects` all over the place in the past. Maybe that has changed?

bil-elmoussaoui:
@dirtydancing i totally agree with you. it would be better to do the opposite

dirtydancing:
Ah, I think I know what is happening here. There is an inconsistency indeed, and it looks like this:

Currently in Numix base via PPA, there is `projects` for 16x16, 48x48 and 96x96. 48x48 is an important size, which is why I recalled `folders`.

In contrast, there is `project` in 22x22, 24x24, 32x32, 64x64, 128x128 and 256x256 (I hope I got that right, but you get the point).

Given this, I would indeed recommend to name all these folders consistently `projects`.

figgyc:
Right. So shall I change project to projects in everything that is project?

dirtydancing:
Yes, I would recommend to do so. And as @wa4557 already pointed out: the place to do so would be this repo: https://github.com/numixproject/numix-folders

bil-elmoussaoui:
@figgyc could you please close this? :)

figgyc:
Done

Foggalong:
We'll copy the folders accross from their own repo to here soon :)

Issue #444, opened 2015-04-09
Touchpad indicator fixes


bil-elmoussaoui:
Hello, 
In the last commit that fixes Touchpad-indicator the icons in the Numix-light are:
- touchpad-indicator-light
- touchpad-indicator-light-disabled
- touchpad-indicator-light-enabled

The icons name must be changed from light to dark.
Thanks!

wa4557:
I'm not so sure about that. What sense does it make to have a dark tray icon if all your other icons are light (or the other way around)? This approach of creating *light.svg and *dark.svg icons is per se a bad idea, and it makes no sense imo to make the same thing in Numix. What is possible and could be done is a symlink from touchpad*-light.svg to touchpad*-dark.svg; but in Numix AND Numix-Light

bil-elmoussaoui:
Well, the icons for the touchpad-indicators are in /opt/extras.ubuntu.com/touchpad-indicator/share/touchpad-indicator/icons
![capture du 2015-04-09 18 22 33](https://cloud.githubusercontent.com/assets/7660997/7071412/675d57a2-dee5-11e4-84da-423a066ace6b.png)
And they uses the touchpad*-dark.svg for the non light indicators.

wa4557:
In the app. Is there somewhere a setting that let you choose between dark and light icons?

bil-elmoussaoui:
Yes! 
![capture du 2015-04-09 18 28 55](https://cloud.githubusercontent.com/assets/7660997/7071554/53847b74-dee6-11e4-9738-7b36e55e6085.png)

wa4557:
In this case my approach is the right one imo. Symlink touchpad-\*-light.svg to touchpad-\*dark.svg in Numix and Numix-Light. Let's see what others have to say

dirtydancing:
Actually, in my view, this kind of issue should be solved in a kind of "third" way: by having **both** panel icons in Numix and **both** corresponding symlinks in Numix-Light.

Indeed, an app that provides light and dark panel/status/indicator icons actually creates a complication. Another example of this I have come across is Variety, cf. e.g. https://github.com/numixproject/numix-icon-theme-circle/issues/1245. Variety lets you choose between a light and a dark panel icon (and provides other choices as well, such as the current wallpaper).

Screenshot of the Variety settings page, current PPA version 0.5.3:

[...]

PR #457, opened 2015-04-13
Indicator light dark

dirtydancing:
Added indicator light and dark for touchpad and for variety, in Numix (icon files and symlinks) and in Numix-Light (symlinks to Numix). Fixes https://github.com/numixproject/numix-icon-theme/issues/444, cf. discussion over there for touchpad and also for variety.


wa4557:
@dirtydancing: Ah ok my bad,  you changed that so that the regular files (e.g. variety-indicator and variety-indicator-light) are now both in Numix-base and both symlinks are in Numix-light. I figured it would make more sense to keep the light variant in Numix-base and the dark variant in Numix-light and make symlinks in each of the both cases.

dirtydancing:
Ah that. Yeah, I figured it would be best to keep both icon files in Numix base, because Numix base already has to be able to serve both icons, depending on the selection of the user.

And usually, when icons are used both in Numix base and in Numix-Light, they are in Numix base and then symlinks are added in Numix-Light. So from my perspective, this is just keeping it consistent.

dirtydancing:
Wait, so are you saying that these icons are not needed in Numix-Light, because Numix-Light inherits from Numix base?

wa4557:
Actually yeah. But on the other hand they don't do any harm neither

dirtydancing:
Because in these cases, the indicator icons are of the some colour in Numix base and in Numix-Light, so the symlinks should not be needed.

dirtydancing:
Good point anyways. Agreed that the symlinks do no harm. Probably this would need some testing to see if they can safely be removed from Numix-Light. Please feel free to go ahead with this if you want to.

dirtydancing:
Otherwise I will look into this. Because there should not be icons/symlinks in Numix/Numix-Light that are not needed. iirc one exception have been the red battery status icons that are the same in Numix base and in Numix-Light, but still symlinks in Numix-Light were needed. But that should be the exception.

Just let me know if you would like to go ahead with this.

wa4557:
Go ahead :) I'm away from computer more or less from now the whole weekend. But yes those symlinks are definitely not needed

dirtydancing:
OK, I will go ahead :)

dirtydancing:
Alas, I ran into some difficulties when testing this with Variety, until I figured out that something is off with Variety itself: apparently, the light icon `variety-indicator.svg` is themeable, but the dark icon `variety-indicator-dark.svg` is not. The Numix icon looks similar to the original icon, so this is not totally straightforward to recognise. But both these icons exist and should be kept in Numix. And yes, the symlink in Numix-Light can be removed, as Numix-Light inherits from Numix base. I checked with the light icon, and it worked. So I will now submit a pull request for this :)

Issue #458, opened 2015-04-14
Remove network-wired from 16 22 and 24/status


wa4557:
I suggest to remove `network-wired` from the `status` folders for sizes `16` `22` and `24` and replace it by symlinking it to nm-device-wired in `devices`.
Reasons:

1. Unnecessary: I tested this using Ubuntu and Cinammon, and it seems to me that they are not needed for displaying the wired network status. This icon is not used at all, and changing it does not change anything in the status bar. @dirtydancing Fancy testing that as well?

2. This icon is used in the Ubuntu-network setting and when using Numix as icon theme it is hardly recognizable (screenshot). If one creates a symlink to ǹm-device-wired.svg` in `devices` it looks way better.
![bildschirmfoto vom 2015-04-14 12 18 06](https://cloud.githubusercontent.com/assets/6475757/7135459/33e2a9ac-e2a7-11e4-980c-9a4f86fa21e0.png)
![bildschirmfoto vom 2015-04-14 12 19 56](https://cloud.githubusercontent.com/assets/6475757/7135460/350a0910-e2a7-11e4-9fee-f42469fa63f6.png)

3. Consistency: `network-wireless.svg` is only present in `devices` and is no monochrome icon. Why treat it differently than the wired icon?

4. According to #189 `network-wired-disconnected` is still missing. This one could also be placed in `devices` and would be consistent with this new approach

dirtydancing:
Sure thing! I will have a look at this. As to 1. "unnecessary": iirc, in the panel `nm-device-wired.svg` is used, so no need for `network-wired.svg` for the panel indeed.

wa4557:
As a side note: `network-wireless-encrypted` also should be moved from `status` to `devices` since it falls into the same category...

wa4557:
Since we are already talking about this icons: Here's a suggestion for an updated nm-device-wired (soon to be network-wired :) icon). The old one was from the elementary-theme and didn't quite fit the Numix style. 
![test](https://cloud.githubusercontent.com/assets/6475757/7143356/47861af6-e2df-11e4-936d-2f457a3f13a7.png)
![test2](https://cloud.githubusercontent.com/assets/6475757/7143355/4784471c-e2df-11e4-8e09-fd351f1e9bd3.png)

Better or worse?

dirtydancing:
I agree with the approach as suggested by @wa4557.

Checked against daily Ubuntu 15.04:

[...]

PR #462, opened 2015-04-16
Removes network wired from status. Adds symlinks in devices

wa4557:
As discussed in #458, removes unnecessary `network-wired.svg` in `status` and adds symlinks to `nm-device-wired.svg` in `devices`


Foggalong:
Presumably the icon would need removing from Numix Light too

wa4557:
Yeah. Good thinking!

Issue #189, opened 2014-10-07
Icons missed on network settings


bil-elmoussaoui:
Hello, i've got a problem on the ubuntu 14.04 using the numix theme and the numix cercle as an icon theme

the issue : 

![selection_001](https://cloud.githubusercontent.com/assets/7660997/4546941/54be78a2-4e48-11e4-81e0-6bd7edd69cf5.png)

some icons are not themed, or i dont know.. 
Thanks

bil-elmoussaoui:
The missing icon is related to #127 
This icon must be in status : 16x16,22x22,24x24,32x32,48x48,256x256

wa4557:
@bil-elmoussaoui I'm not sure about the status thing. I think this icon should be in devices. Not in status

PR #467, opened 2015-04-19
nm-device-wired enhancement, adds network-wired-disconnected

### LINKED ISSUE #461: Icon enhancement for nm-device-wired.svg in devices
In continuation of https://github.com/numixproject/numix-icon-theme/issues/458 and https://github.com/numixproject/numix-icon-theme/issues/460: the issue of an icon enhancement for `nm-device-wired.svg` in the **devices** directories (not in `status`) has come up. Currently, this icon looks somewhat inconsistent with the other Numix icons. Example screenshot from `48x48/devices`:

48x48_devices

Alternative designs (if an alternative design is chosen) could display the network cable plug like currently in `scalable/status` and in `48x48/notifications`, or the actual plug socket.

wa4557:
Fixes #461 all credit to @dirtydancing who made the icon :) This pull-request only makes sense if #462 is merged as well


dirtydancing:
@wa4557 Thanks for the credit! This was a collaborative effort, without your base draft no icon :)

One question: do I understand correctly that you deleted `nm-device-wired.svg` in sizes larger than 48x48? If so: this should not be done. We need the symlinks `network-wired.svg` to the target `nm-device-wired.svg` for all sizes up to 256x256, already because of the Arch Avahi icon problem. Currently, `nm-device-wired.svg` is present up to 256x256 as well. And with `network-wired.svg` (symlink) and `nm-device-wired.svg` (target) present up to 256x256, I guess the corresponding `network-wired-disconnected.svg` can be present up to 256x256 as well. What do you make of this?

wa4557:
Hmm. But for the avahi problem only 48x48 sized icons are needed. All the other similar icons (`nm-device-wired`, `network-wireless` etc.) stop at 48x48. Most of the `devices` icons turn into monochrome-icons at 64x64 (and those who don't are not original Numix icons) therefore if we would add `nm-device-wried` (and `network-wired-disconnected`) they should be monochrome too

dirtydancing:
The Avahi problem does not stop at 48x48. E.g., in the GNOME shell app overview, the icons displayed are larger than 48x48, cf. e.g. this Antergos screenshot:

55ff3340-e51b-11e4-9a0b-4a9745cff418

I am aware that most icons in `devices` turn monochrome at 64x64, but there are several exceptions to that, and so far `nm-device-wired.svg` is one of those icons that do not turn monochrome up through to 256x256. So this would just be continuing what is already there :)

dirtydancing:
That being said, I see your point, and maybe we can get away with 48x48 also for the Avahi problem. Let me check that. 48x48 in any case is much better than currently, with `network-wired.svg` only extending up to 24x24.

wa4557:
Well all the app icons are 48x48 as well

dirtydancing:
Yeah, but the app icons are also in the "scalable" directory in Numix-Circle (symlink to 48x48/apps), but the icons in Numix base are not. Theoretically, also the current `network-wired.svg` in 24x24 is a scalable .svg icon, but it does not get scaled well in GNOME Shell at all. That is why I will check with Antergos.

dirtydancing:
And here comes the evidence: indeed, when the icon is added in 48x48, it still does not scale sharp, in contrast to the screenshot above where the Avahi icons are sharp.

antergos_avahi_48x48

Given this evidence, I would suggest to add these up to 256x256. They do not break anything there, but they fix this problem (and maybe others?). What do you think?

wa4557:
Fine, I'll add them :)

dirtydancing:
Cool, thanks! :)

dirtydancing:
Now we are talking sharpness :-) Screenshot of Antergos with the new icons in present in all sizes:

antergos_avahi_new

wa4557:
Cool, although I still think that ahavi should solve that issue, not Numix

dirtydancing:
Yeah, I guess we all agree on that. But it does not break anything to have these icons in place in the bigger sizes as well.

palob:
@dirtydancing @wa4557 There are a few small alignment issues with the icon. Please align all straight horizontal/vertical lines as much as possible to the grid (in some cases for all sizes seperately).

I like the design :+1: and the old design could very well do with an refurbishment but (without testing) isn't it too busy detailed for the small sizes?

wa4557:
I'll do my best. In some cases however it is not possible without destroying the design too much (especially for smaller sizes where the aligning is more crucial then for bigger sizes).
Frankly, the smaller sizes won't be used a lot (the smalles icon I'm aware of that will be used is 32x32 the smaller ones are just there for consistency) therefore I don't think it is necessary to design a new icon just for smaller sizes that won't be used

dirtydancing:
@palob Thanks for looking into this! I will do some testing and see what gives (might be a little while).

@wa4557 It might be that the Ubuntu network settings use size 24x24 of `network-wired.svg` (symlink).

dirtydancing:
OK, I have done some testing, and @palob has raised a valid point: there is some difficulty with the small sizes. So I have come up with a "simplified" version, which in my view turned up promising enough to actually work with all fixed sizes. @wa4557 would you mind if I open a new branch/pull request for those and to still let this one open, so we can compare? I am interested in giving the different sizes a shot in Inkscape :)

Anyways, before I get into this, I would appreciate your views on the new draft, so if you fancy to have a look, here we go: screenshots with enlarged size 48x48:

Current version (connected and disconnected):

current

current_disconnected

New draft version (connected and disconnected):

draft_new

draft_new_disconnected

dirtydancing:
To provide some screenshot about what I mean: E.g., Ubuntu is using a 24x24 icon in its network settings.

Current Numix-Circle:

Connected (corresponding issue report: https://github.com/numixproject/numix-icon-theme/issues/458):
1a_current_connected

Disconnected (corresponding issue report: https://github.com/numixproject/numix-icon-theme/issues/189):
1b_current_disconnected

Proposed here connected:
2a_proposed_connected

Proposed here disconnected:
2b_proposed_disconnected

New connected:
3a_new_connected

New disconnected:
3b_new_disconnected

palob:
I think I prefer the simpler design + the distinction connected/disconnected is bolder.

wa4557:
@dirtydancing Please go ahead and do this  I'm quite busy this week :) I like the new icon as well (although the colors in the connected icon are a bit harsh imo). I'm closing this then

dirtydancing:
Agreed that at least for the smaller sizes the simpler design seems to fit better. I had another look at the size 24x24 icon and ironed it out a bit (was a bit too big, and at some points not completely properly aligned). As a result, the wireless at least looks a little bit less out of line :)

new_connected

new_disconnected

Possibly take the simpler design for the smaller sizes and the more detailed one for the larger sizes? The colours match up (just a tad darker outside grey for the simpler ones to that they match up).

dirtydancing:
@wa4557 How come that we are commenting at the same time so often? Once more just happened here :) OK then, I will go ahead with the simpler design.

wa4557:
Cool. Could you play a little bit with the colors, making the green a little bit less turquoise, it is a little bit too much blue for me? Anyway the icon looks cool as it is

palob:
Yeah, maybe play a little with the colours on the connected one :)

dirtydancing:
How about just having the green in the connected one and not the yellow. Green stands for linked up and should be enough of a contrast with the dark grey for disconnected. What do you think?

connected
new2_connected

disconnected
new_disconnected

palob:
I think it was better to have the entire inner part coloured distinctly different.

dirtydancing:
Yeah, agreed that the distinction was better before, and the icon was more interesting, too. I will think of s.th. and go ahead then. Thanks for all your feedback!

Issue #460, opened 2015-04-15
Icon nm-device-wired.svg is inconsistent in 22x22/devices


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/issues/458, because this is a separate issue: The icon `nm-device-wired.svg` in the **devices** directories (not in `status`) currently is a "golden" non-monochrome icon. But in `22x22/devices`, there is an inconsistency and the icon looks like this (apparently outdated panel design from before the fixed size status icons redesign):

![devices_nm-device-wired](https://cloud.githubusercontent.com/assets/7164227/7155311/0c0bc3b6-e35e-11e4-9492-917212fad45a.png)

Foggalong:
@dirtydancing I think this icon might be a status design but located in devices because of Xubuntu

dirtydancing:
Yeah, this one might have been used in Xubuntu, but it is not anymore. iirc I came across a very old screenshot of Xubuntu that displayed this one, but I have not seen it for a long time. Should be save to "remove", i.e. make consistent with the other fixed sizes.

Foggalong:
Ah that's good to hear. If I merge [this pull request](https://github.com/numixproject/numix-icon-theme/pull/468) will it take care of this issue too then?

dirtydancing:
Yes.

Foggalong:
\O/

PR #468, opened 2015-04-21
Enhanced and added wired icons

dirtydancing:
Enhanced `nm-device-wired.svg` and added `network-wired-disconnected.svg` to all fixed size devices directories. Fixes https://github.com/numixproject/numix-icon-theme/issues/461, https://github.com/numixproject/numix-icon-theme/issues/460 and https://github.com/numixproject/numix-icon-theme/issues/189.

This pull request only makes sense in combination with https://github.com/numixproject/numix-icon-theme/pull/462 (adds symlinks `network-wired.svg` with the target `nm-device-wired.svg`).

@wa4557 @palob In continuation of https://github.com/numixproject/numix-icon-theme/pull/467 here come the new icon versions.

connected (enlarged 48x48)
connected_48x48

disconnected (enlarged 48x48)
disconnected_48x48

Overview of the different sizes:
wired


wa4557:
@dirtydancing Good job I really like them (especially for the small sizes they fit better). I wonder if for the bigger sizes the (from 32x32 upwards) the more detailed would be more fitting (especially when looking at the other icons in `devices`). That being said, I would also merge them like this :)

dirtydancing:
@wa4557 Thanks! I will do some testing a little later and provide some screenshots, to visualise.

dirtydancing:
OK, screenshots of Ubuntu network settings (starting point for these icons) and of Antergos for the Avahi problem (which can be "fixed" as a side effect, but as we all agree should be fixed upstream).

Ubuntu, connected and disconnected (all fine)

connected

disconnected

Antergos (all fine, as much as it can be fine with a Numix base icon)

antergos_new

Antergos with "most used" app overview (all fine)

antergos_new_2

Given these screenshots, I think I prefer to go with one consistent icon here, and not bring in the more detailed one for the larger sizes, so the same icon will be displayed no matter which size might be used.

palob:
Well, one last idea :)
How about having the inner part dark altogether for the disconnected path in a way that it really screams out **DISCONNECTED**?

dirtydancing:
Interesting idea, let's see :) Here is the alternative with an all-dark inner part for disconnected:

24x24 icon enlarged:

alternative_24x24

In Ubuntu network settings:

alternative

Seeing this, in comparison I think I prefer the two inner colours as proposed above:

9869131c-e86e-11e4-877c-9b43fff1e6f2

To me, the contrast to the outer "frame" is not as good with the completely dark inner part (the contrast is a close call in this case, though). It also does not scream out, at least to me, but rather makes the icon "heavier" (and making the inner part even darker for still better contrast would make it even heavier). Moreover, these icons are designed to work with light and dark backgrounds, and having only dark colours would tend to work only on lighter backgrounds. It also does not correspond as well to the connected icon, which also has two colours in the inner part. And finally, I do not see how these two icons could be confused :)

95055802-e86e-11e4-9630-dd03fdff9b2f

What do you make of this?

dirtydancing:
Some more screenshots, this time relating to Xubuntu's notification for wired connected.

As already mentioned, for the disconnected notification Xubuntu (15.04 Beta 2) is using `notification-network-ethernet-disconnected.svg` from `48x48/notifications`:
disconnected_notification

But, for the connected notification, instead of using `nm-device-wired.svg` from `48x48/notifications` (like Ubuntu):
connected_notification

Xubuntu is using `nm-device-wired.svg` from `48x48/devices`:
connected_current

With the new design from above it would look like this:
connected_new

In comparison here is the more detailed design variant:
connected_detailed

Also in this case, I think that the less detailed design currently suggested here is the most fitting one. It is also consistent with the design of the Numix notification icons and thus very versatile.

b4a05cc2-e84e-11e4-80b0-732cfc5d2090

dirtydancing:
To complete the round of screenshots, here is Antergos with Circle and with ootb default Square, to show that the icon suggested here fits in with both icon themes as well as with common wallpapers:

Numix-Circle, Antergos default wall (all fine):
circle

Antergos ootb default Numix-Square, GNOME 3.16 default wall (all fine):
square

Bonus: just for demonstration, here is a screenshot of Antergos with the Technic icon theme (currently in development, this is out of the Numix labs from @Foggalong). I figure that this icon theme might resonate with users of Arch & Co., so this will pose a problem with ootb Arch as well, just to let you know:
technic

dirtydancing:
To clarify: although I am presently not an Arch user, the Technic icon theme resonates with me as well :)

palob:
As said before I think this isn't an issue Numix should care about. 
It's a GNOME usability bug that there isn't a comfortable way to edit your launchers from within the shell overwiev + it's an Arch annoyance bug (which they won't fix due to the Arch way) that it installs all that launcher clutter.

It could however be added to the hardcode fixer script, fixing the launchers to use an actual app icon.


I think the proper way (provided you don't need them) to fix this is to hide the launchers altogether by adding the line ``NoDisplay=true``.
The trick is to copy the launchers to **~/.local/share/applications** :) Your local launchers override the global ones and this way the Avahi stuff won't reappear after every update.
If you do need the launchers you can assign a "durable" custom icon to the launchers this way as well.

You might have to restart GNOME Shell for your local launchers come into effect.

dirtydancing:
Agreed that the Avahi issue is not a Numix issue. Yet, if the opportunity presents itself, there seems to be no reason not to address this breakage, esp. with Antergos using Numix-Square as default icon theme.

dirtydancing:
As we all agree that the "Arch Avahi issue" is not a Numix issue, I guess the question is: Would this icon be designed differently if the Arch Avahi issue did not exist? I am not sure that it would; any suggestions?

Additionally, in Ubuntu 14.10, `network-wired.svg` is also used as a general icon for the network connection, no matter if wired is connected or disconnected.

Screenshot with the proposed changes (when disconnected):
network_1

This icon currently is hardly visible, but it is there:
network_2

So in addition, I am wondering if it might be confusing that "connected" and "disconnected" are showing next to each other? Would that then be an Ubuntu issue?

And how about the Xubuntu notification icons as described above? That is a different notification system from the one in Ubuntu. Is that then a Xubuntu issue?

You probably see what I am getting at: so many "issues" around here, with all kinds of icon sizes needed in all kinds of different situations. I have just mentioned those that I am aware of, there might be more.

So how do deal with this from Numix's side (if at all), when Numix is catering to all distributions? :)

dirtydancing:
For comparison, I have checked a design variant that is closer to the present notification icons.

I think I still prefer the "coloured" icons suggested above, as to me this variant is not an improvement.

device/"connected":
connected

disconnected:
disconnected

Result in Ubuntu 14.10 network settings:
network_3

dirtydancing:
Actually, on a second look, this even "simpler" design variant has s.th. going for it as well :). It has a kind of "neutral" device/"connected" design, and the red cross is clearly showing a disconnected status. Moreover it also fits in quite well e.g. with the tower of `network-wireless.svg` above it. Any thoughts?

ubuntu1410_network

dirtydancing:
And of course, the colours can be slightly modified to correspond to already existing colours in Numix:

Numix white `#f9f9f9` for the center (instead of #f1f9f9) and `#909ea9` for the outside (instead of #a3adad), which is used in `devices`, esp. for the larger sizes. Red is already "warning red" `#ef555c`.

Result:

connected

disconnected

wa4557:
I like the new ones (although I think a little more details like the pins wouldn't hurt). Concerning notifications: What do you think about removing the notification icons (except sound and brightness) and instead symlink the corresponding icons from `devices` to `48x48/notifiactions`? I suppose they would fit better

dirtydancing:
OK, here comes the new design, @wa4557 good call about the pins, included those :). Screenshots:

connected

disconnected

wired_2

ubuntu1410_network

ubuntu1410_network_2

dirtydancing:
@wa4557 Not sure about what you suggested about the notification icons. Could you possibly open a new issue for that, as otherwise things might get too mixed up here :) Maybe it also makes sense to see first if and how this pull request gets merged, and then to have another look at the notification icons.

wa4557:
even bester now! Yeah one thing fter aanother. I'll eventually open a new issue later :)

dirtydancing:
Improved some small size icons to be more consistent. Result:

wired_3

wa4557:
Remove white background? Seems to better without

dirtydancing:
Removing the white (currently Numix white `#f9f9f9`) background and just leaving it transparent does not work e.g. when selected. The contrast is somewhat low then, and the red X for "disconnected" is not visible anymore.

Screenshot of Ubuntu 14.10 network settings with Numix GTK theme:

transparent

palob:
Still tweaking :)
As for the pure white design, I think the pins conflict with the cross.
If you prefer the white design now how about centering the red cross + no "pointy edges" (just an **X**)?

dirtydancing:
Actually, when I checked centering the red X before, I did not like it. But never say never :)

Anyways, the point that @wa4557 made about the transparency in the center lead me to check a dark variant, given that transparency is not an option here. What do you make of this "coloured" variant, which also in a way comes back to @palob suggesting a "dark" center earlier on. iirc correctly Numix black is `#292929`, so I used that one, and the "X" is better fitting in white here, so it is Numix white `#f9f9f9`. 

I like this one, and since this icon does not have to be monochrome, this should be possible:

connected

disconnected

black-white

dirtydancing:
Quick feedback @palob concerning the "X" interfering with the pins: indeed, that is once again a valid point :). I will have another look at that. Two quick screenshots to give a rough idea:

I am still not convinced about centering:

cross_1

but there might be s.th. to having no pointy edges (as currently in the notification icons):

cross_2

dirtydancing:
So here we go: new and improved design. The "X" is difficult to handle in the small sizes, but the "X" actually is not needed at all. This design does the trick, surprisingly simple and efficient. I prefer this one:

connected

disconnected

wired_4

ubuntu1410_network-settings

ubuntu1410_disconnected

ubuntu1410_connected

wa4557:
Looks good as well although a bit dark:) this Icon arguably might the one with the most design changes :P

dirtydancing:
Well, the design changes are not only aesthetic ones, but also due to a functional evolution of this icon: the more detailed one did not work in the (most relevant) small sizes. The coloured one was leaning somewhat to a status ("connected") icon, neglecting a little the "neutral" device part of the very same icon. And the "X" in the white version did not work properly, esp. in the small sizes. So here we are :)

wa4557:
Yeah I know, it's just funny how it changed from the first draft :) Well I think these icons should work pretty well, so I think they should/could be merged. I still have a few things on my list that might be imrpoved in the base theme, but I think if all the pull-request are merged at least the hardest things are done :P

dirtydancing:
Agreed :) And although that is not the focus of this icon (and should not be, as that is not a Numix issue), even the "Arch Avahi issue" would be mitigated along the way, as this Antergos screenshot shows:

antergos_square

Foggalong:
Kudos on going with [these](https://cloud.githubusercontent.com/assets/7164227/7334552/2e5f39c0-eb97-11e4-8fe0-b1dc19d9b6d7.png) as the final design without any dev intervention because it was by far the best design, and props on having everything aligning to grid too.

dirtydancing:
Thanks @Foggalong, it has taken a couple of sessions, but we got there :)

Issue #438, opened 2015-04-08
Remove bluetooth.svg from Numix base, keep in Numix-Circle


dirtydancing:
In the course of testing different inheritance possibilities, cf. https://github.com/numixproject/numix-icon-theme/issues/190#issuecomment-90890423, I came across an inconsistency with the Bluetooth icon: `bluetooth.svg` currently is a symlink in Numix base, pointing to the target `bluetooth-active.svg`. 

But the icon names for the Bluetooth status are `bluetooth-active.svg` and `bluetooth-disabled.svg` , so `bluetooth.svg` is not needed here.

To make things worse, `bluetooth.svg` is displayed in the system settings overview in Ubuntu. This should be a consistent Circle icon if Numix-Circle is selected (`bluetooth.svg` exists in Numix-Circle and should be kept there), and it should be the original icon in the system settings overview if Numix base is selected, not the `bluetooth-active.svg` status icon (target icon) from Numix base.

In conclusion, I would like to suggest to remove `bluetooth.svg` from Numix base.

@wa4557 If you fancy to take a look: Can you think of a reason to keep `bluetooth.svg` in Numix base?

wa4557:
@dirtydancing No. In fact if we keep it and go along with the renaming and `apps` folders approach, this icon should be renamed, so if it is deleted or renamed doesn't make a lot of difference...

dirtydancing:
@wa4557 Agreed. Actually, I checked this with your testing branch of numix-icon-theme, the one that already has the renamed icons in it (and the `apps` folders), and `bluetooth.svg` is still present there (not renamed yet).

Given the choice for this icon between renaming and deleting, I think I would go for deleting, as `blueetooth.svg` as an "app status icon" appears not to be needed in Numix base at all (also not "re-renamed").

The only reason for keeping it and renaming it rather than deleting might be that another distribution could be using this specific icon for the panel (?). So I guess renaming would be fine as well.

Here are screenshot with Numix base and Numix-Circle, both via PPA, to demonstrate what I mean with the system settings overview:

Numix-Circle (all fine)
![numix-circle](https://cloud.githubusercontent.com/assets/7164227/7045559/7ef416ee-ddff-11e4-9b81-9229b5f0b08b.png)

Numix base (current)
![numix](https://cloud.githubusercontent.com/assets/7164227/7045582/bbc80da0-ddff-11e4-8e7a-94df2c1a81cb.png)

Numix base (proposed without `bluetooth.svg`)
[...]

Issue #192, opened 2014-10-10
Some icons now appear pixelated on Gnome 3.14


spacepluk:
Like this:

![screenshot from 2014-10-10 08 52 03](https://cloud.githubusercontent.com/assets/824762/4588480/f6aa3bb8-5043-11e4-85f6-ae4b458e836a.png)

spacepluk:
This bug is back on 3.16.

wa4557:
Try the version from github instead of the one from PPA. Should be fixed because all the status icons moved to numix-base

spacepluk:
I'm using AUR's package in Arch Linux which takes the latest version from github.

Foggalong:
The icons being in Numix base is exactly the problem here. Because base theme doesn't have any appnicons is going the same named panel icons and using them for everything. Ideal would be to take

wa4557:
Or merge base and circle or add an inbetween numix-mono as @dirtydancing suggested :)

dirtydancing:
@wa4557 Please cf. here https://github.com/numixproject/numix-icon-theme/issues/107#issuecomment-60065491 for the difficulties that an additional numix-mono would entail :-)

Merging base and circle leads to the problem (in this context) that e.g. for those apps that do not have a distinct app status icon name (e.g. Google Chrome and VLC), those "status icons" would have to be deleted, whereas they should not be picked up from Numix base because Numix-Circle inherits from Numix base and so always the Circle app icon from Numix-Circle should be picked up, without recourse to Numix base.

(Of course, merging base and circle e.g. would also mean that also Square and base would have to be merged, and non-uniformity and base, so all the modularity with the different Numix icon themes would get lost. But that is a more general point.)

Then again, this thread seems to suggest that the app "status icons" for Google Chrome and for VLC are still picked up with gtk 3.16, even though they are in base? That would need further inquiry.

@spacepluk Could you provide a current screenshot, if possible? We need to assemble evidence here and possibly try and reproduce this :-)

Currently, I believe that the symlink approach suggested in https://github.com/numixproject/numix-icon-theme/issues/190 ist the most promising (if that is possible at all). Probably this conversation should be continued over at https://github.com/numixproject/numix-icon-theme/issues/190 anyways.

spacepluk:
Sure :) there you go:
![rorb4iv](https://cloud.githubusercontent.com/assets/824762/6936766/356c4c90-d85e-11e4-8d22-b018392d3351.png)

[...]

PR #471, opened 2015-04-27
Add missing blueman status icon

hashworks:
I forgot one in my pull request a few weeks ago.

### DIFF
--- a/Numix-Light/16x16/status/blueman.svg
+++ b/Numix-Light/16x16/status/blueman.svg
@@ -1,0 +1,1 @@
+bluetooth.svg
--- a/Numix-Light/22x22/status/blueman.svg
+++ b/Numix-Light/22x22/status/blueman.svg
@@ -1,0 +1,1 @@
+bluetooth.svg
--- a/Numix-Light/24x24/status/blueman.svg
+++ b/Numix-Light/24x24/status/blueman.svg
@@ -1,0 +1,1 @@
+bluetooth.svg
--- a/Numix/16x16/status/blueman.svg
+++ b/Numix/16x16/status/blueman.svg
@@ -1,0 +1,1 @@
+bluetooth.svg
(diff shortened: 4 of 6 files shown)


dirtydancing:
Are you sure that the icon/symlink `blueman.svg` is needed as a status icon in Numix base?

I would like to draw your attention to https://github.com/numixproject/numix-icon-theme/issues/438. Anyways, `bluetooth.svg` is a symlink itself, with the target `bluetooth-active.svg`, so the target for this symlink should be rather either `blueman-active.svg` or `bluetooth-active.svg` instead of `bluetooth.svg`.

What do you make of this?

hashworks:
Originally, there are two seperate `blueman.svg`:
```
/usr/share/icons/hicolor/scalable/apps/blueman.svg
/usr/share/icons/hicolor/scalable/status/blueman.svg
```

The missing status icon resulted in blueman some times switching to the colored icon you see in #438. Adding a symlink for `status/blueman.svg` solved this. Not sure why the status icon interferences with the app icon.

dirtydancing:
Ah, I see. Maybe `blueman.svg` is different from `bluetooth.svg` in `status` then (as far as I can tell, `bluetooth.svg` is not needed in `status`).

I am afraid that this is part of a bigger and more complex problem: that of app icons and their corresponding status icons having identical names. This might work out with the icon lookup if one single icon theme is concerned, such as `hicolor`. Numix-Circle comes before Numix base in the inheritance line (Numix-Circle inherits from Numix), and problems can arise with identically named icons due to that, or in general. Alternatively, a problem might arise due to the different sizes of these icons, and depending on how large the displayed icon is, the design will change. This whole problem is somewhat confusing, and it does not make it easier that originally these two icons were in the `scalable` directories. Bottom line: I figure it is simply bad practice to name these differently designed icons identically.

If you are very interested in this subject, please cf. https://github.com/numixproject/numix-icon-theme/issues/192 and https://github.com/numixproject/numix-icon-theme/issues/190 and prepare for a long read :)

@wa4557 Are you still keeping a list of icons that would need to be renamed? If this is merged, I figure `blueman.svg` would be one more :)

PR #477, opened 2015-05-02
Wrong size of network-vpn in 48

wa4557:
This one slipped through in my previous pull-request


Foggalong:
Some of them are also showing up plain grey :/

wa4557:
Yeah but that's intentional. All of the devices icon become grey from size 64 upwards

Foggalong:
Of course - my bad :P

Issue #459, opened 2015-04-15
Panel bar sound icon smaller than the others


pedrofleck:
I'm using Arch with the git package from AUR and Gnome 3.16. The volume icon in the panel bar is smaller than the power and network icons making it low res and ugly (screenshot attached), how can I fix it?
![captura de tela de 2015-04-14 21-43-58](https://cloud.githubusercontent.com/assets/761321/7150299/635bb29e-e2ef-11e4-86cf-09f74220b8cf.png)

dirtydancing:
Hi, are you sure that this is not the Numix icon? What should it look like in your view? I am asking because this looks like the Numix icon to me.

The volume icon in your screenshot should be `audio-volume-medium-symbolic.svg` from the`scalable/status` directory. Screenshot of that icon in Numix:

![numix](https://cloud.githubusercontent.com/assets/7164227/7155060/ee410e56-e35b-11e4-9041-1db1bd872169.png)

In contrast, at least in the version of gnome icon theme shipping with Ubuntu 14.10, that icon in gnome looks like this:

![gnome](https://cloud.githubusercontent.com/assets/7164227/7155063/0296b0ea-e35c-11e4-916e-c1dd8cc7d8e3.png)

freckts:
Yeah, you're right, but why the icon is smaller? It happens with every GS theme and was present in 3.14 too.

dirtydancing:
Was it different before GNOME 3.14? The volume icon in your screenshot might just look smaller because the `network-wired-symbolic.svg` icon next to it looks somewhat "massive" in comparison. This is a screenshot with both icons next to each other, and this looks just like in your screenshot. One possible way to address this would be e.g. to redesign the `network-wired-symbolic.svg` icon:

![icon-sizes](https://cloud.githubusercontent.com/assets/7164227/7160820/be532280-e38b-11e4-8796-4edc18ec23ad.png)

dirtydancing:
To illustrate, here are some screenshots of Ubuntu GNOME 15.04 beta 2 with default GNOME Shell 3.14.4. As you can see, with Numix icon theme the volume icon looks quite small in comparison indeed:

GNOME icon theme:
![gnome](https://cloud.githubusercontent.com/assets/7164227/7161959/47ecf6e6-e392-11e4-955b-a5d3260ae199.png)

Adwaita icon theme:
[...]

PR #478, opened 2015-05-02
Fixes #473

### LINKED ISSUE #473: Enhance notification icons
Issue concerning all *buntu distributions.
This issue only makes sense if most of the current pull-requests (esp. #468) are merged and #472 is addressed:
The current notification icons are basically scaled up status icons. I think it would be better if icons from `devices` would be used, because that would mean that they are less grey and dull. An excpetion for that would be the sound notification and brightness notifications because they are somehow special considering their placement and greyish progress bar.

So my idea would be something like this:

bildschirmfoto vom 2015-04-29 14 17 39

- Symlink battery icons from devices to notifications (only works if and when #452 is merged). 
- Symlink `network-wireless` in `devices` to `nm-device-wwan` in `notifications`. 
- Symlink `nm-device-wired` in `devices` to `nm-device-wired` in `notifications` (the icon will probably look different in the future, see #468).
- If #472 is addressed symlink `nm-device-wireless`in `devices` to `notifications` (if #472 is not addressed do that as well, but change the design of `notification-network-wireless-disconnected`)
- If #468 is merged (that's the reason the symlink is broken currently) symlink `network-wired-disconnected` to `notification-network-ethernet-disconnected` and 
- design two new icons for `notification-network-wireless-disconnected` and `notification-gsm-disconnected` (my suggested icons can be seen in the screenshot). 

This generally leads to nicer notifications and solves the bug, that Xubuntu takes `nm-device-wired` and `nm-device-wireless` from `devices` instead of `notifications` for its notifications

wa4557:
Since we have now new and pretty devices icons the notifications icon for ethernet needs to be made consistent. Fixes #473 (kind of :)

### DIFF
--- a/Numix/48x48/notifications/notification-network-ethernet-connected.svg
+++ b/Numix/48x48/notifications/notification-network-ethernet-connected.svg
@@ -13,7 +13,7 @@
    id="svg2"
    version="1.1"
    inkscape:version="0.91 r"
-   sodipodi:docname="nm-device-wired.svg">
+   sodipodi:docname="notification-network-ethernet-connected.svg">
   <metadata
      id="metadata12">
     <rdf:RDF>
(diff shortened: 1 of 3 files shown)


Foggalong:
Has the other ethernet design been completely remove from everywhere else in the theme then?

wa4557:
Almost. The symbolic icon has still the old design. Replacing the symbolic icon with a variant of this one would also fix #459 btw

Foggalong:
Ah, feel free to add those to this request then. Kill two birds with one stone :)

wa4557:
Will do

wa4557:
Did

dirtydancing:
@wa4557 As to the "symbolic"-named icon: Please note that the 16x16 icon has been designed for the fixed size 16x16 directory, not for the `scalable` directory. The icons in the `scalable` directory are somewhat different, mostly larger in relation to the canvas (e.g. also for the battery icon). So I am wondering if this design is fitting for the "symbolic"-named icon in `scalable` (from what I can tell, the design form has stayed the same?).

Maybe we will need to address this and make it more consistent with the other "symbolic"-named icons. It not checked yet, this might need testing in GNOME Shell, which is where this will be displayed.

dirtydancing:
Here are two comparative screenshots to demonstrate what I mean:

Design before:
before

Design current:
current

As you can see, the icon is indeed to small for this purpose and breaks consistency. Given this, I would recommend to revert this change (for this single icon). The size inconsistency (not too bad imo) in GNOME Shell panel rather is a matter of the volume icon being too small. To clarify, I do think that the direction of this redesign is better also for GNOME Shell, it just needs some more attention.

wa4557:
Yeah you're right, obviously. But it's just one icon, so another pull-request should solve that issue. Do you want/can do that? I'm a bit busy right now

dirtydancing:
This will require some more time, because this involves the consistency of the GNOME Shell panel. I am happy to have a look into it (also into the volume icon), but this will take some time. So for the time being, I would recommend to revert this one.

wa4557:
Cool I'll create a pull request that reverts this change in scalable.

PR #480, opened 2015-05-03
Reverts network-wired in symbolic

wa4557:
It seems that the new `network-wired-symbolic` design doesn't fit yet (cf. #478 comment/screenshot by @dirtydancing) therefore here a PR that reverts that commit. 
But @dirtydancing promised to look into this issue, so he will eventually come up with a new icon for the scalable directory :)


dirtydancing:
Actually, I have come up with a suitable icon already. So @wa4557 was right: possibly no need to revert this, but rather go ahead with the new icon right away. Here it is:

Icon before the last change (the changed icon does not fit, so I am leaving that one out):

1_before
2_before

Proposed icon (consistent with the other "wired" icons in Numix base as well as with the GNOME Shell icons):

1_proposed
2_proposed

Thoughts? @Foggalong Should I open a new pull request for that? This one could be closed then.

wa4557:
Good one! How does it look like when 1px wider? It looks a bit compressed

dirtydancing:
Ah, I just noticed that `network-wired-symbolic.svg` has been changed in Numix, but not in Numix-Light. So I will go ahead and create a pull request anyways. This pull request can be closed then depending on what is decided.

dirtydancing:
Cross-posted again :) Let's get into a discussion on the new proposed icon over in the new pull request.

wa4557:
Fine

PR #482, opened 2015-05-03
Enhanced network-wired-symbolic.svg

dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/pull/478 and https://github.com/numixproject/numix-icon-theme/pull/480, this is an enhanced version of `network-wired-symbolic.svg` for Numix and for Numix-Light.

This icon is consistent with the other "wired" icons in Numix base, and it is consistent with the other Numix icons in GNOME Shell.

Icon:
proposed

In GNOME Shell (3.16):

1_proposed
2_proposed

### DIFF
--- a/Numix-Light/scalable/status/network-wired-symbolic.svg
+++ b/Numix-Light/scalable/status/network-wired-symbolic.svg
@@ -1,3 +1,61 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="16" viewBox="0 0 16 16" height="16">
- <path d="m 0,1 0,9.999909 3,1.82e-4 L 3,13 l 2,0 0,2 6,0 0,-2 2,0 0,-2.000091 3,0 L 16,1 z M 3,3 13,3 13,6 3,6 z" fill="#353535"/>
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   viewBox="0 0 16 16"
+   height="16"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.91 r"
+   sodipodi:docname="network-wired-symbolic_LIGHT.svg">
+  <metadata
+     id="metadata10">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs8" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1215"
+     inkscape:window-height="776"
+     id="namedview6"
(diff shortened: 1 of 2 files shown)


wa4557:
Hmm, why not decrease the height 1px to make it vertically centered + it doesn't look so compressed anymore
unbenannt

dirtydancing:
@wa4557 I created this one with the size 14x15 on a 16x16 canvas because 16x16 seemed too large and 14x14 seemed too small. To me, maybe on very first sight only, this looked a little compressed as well, but only because I compared it to the present Numix icon, which is very wide. I have not checked 15x15 on a 16x16 canvas, possibly because I thought this might be off-center then (that one is 16x14).

dirtydancing:
14x14 will look small when compared to the battery icon. Also, it will not be on the same lower line, so actually 14x15 looks more aligned :)

wa4557:
Ok and making it 16x15? Aligned to the bottom?

dirtydancing:
Too wide :) Let me get some screenshots so we can compare.

palob:
@wa4557 Just a side note in case you didn't know: it is better to move upt the nodes of the lower part by 1px instead of decreasing the height of the icon by 1 px.
This prevents a blurry / misaligned lines :)

Did the previous icon exceed the 16x16 canvas?

dirtydancing:
No, the previous one was 16x14.

dirtydancing:
Previous:
previous

Proposed:
proposed

palob:
How about 16x14 for the new one?
(Sorry if that's rubbish for some reason. I didn't follow closely.)

dirtydancing:
@palob No no, we are just starting this conversation :) 16x14 to me looked too compressed.

Is there anything that prohibits using an odd number like 14x**15** in general? Just wondering.

dirtydancing:
@palob Ah, you meant 16x14, so 16 wide 14 high, just like the previous one? I have not checked that so far.

dirtydancing:
Just as a direct comparison of previous and proposed:

1_before
1_proposed

2_before
2_proposed

dirtydancing:
Question: if you look only at the proposed version (not in comparison to the previous one), does the network-wired icon look compressed to you?

palob:
Yeah, illusionarily it doesn't look like a square to me.

dirtydancing:
Well, it is not a square :) And to me, it does not have to be a square. A square is just how it is often handled. But of course, this is not supposed to cause irritation either.

@wa4557 possibly we will still have to revert then for the time being, as currently this is broken in Numix base. Well let's see if we can come up with something here that we can agree on.

palob:
Ahh yes, it's taller than wide indeed :) 
But in case the horizontal padding isn't needed and it won't look weird why not 16 px wide?
In real life the network plug is wider than tall.

wa4557:
Better idea: one can move the grid half a pix to the right/left (@dirtydancing : Dokumenteinstellung Gitter :). By doing so you could make a 15x15 that is aligned to the grid

palob:
Well, Inkscape's grid is only a means of assistance. Shifting it won't have any influence on the blurry outcome.
The real pixel grid still will originate from (0,0).

dirtydancing:
First things first :) Here is a screenshot with a centered 14x14 icon:

14x14

dirtydancing:
Do not pay too much attention to the size of the volume icon, that one is somewhat small in comparison to the others, cf. also https://github.com/numixproject/numix-icon-theme/issues/459.

wa4557:
I prefer 14x14
@palob You sure? Looks sharp to me when moving the grid

palob:
The grid boxes represent your display's pixels :)
How would you be able to shift the pixels?

wa4557:
Ah yes quite stupid, you're right of course :blush:

dirtydancing:
16x14:

16x14

dirtydancing:
16x16:

16x16

palob:
16x16 seems to be in line with the battery icon. But wasn't that to be redesigned (downsized) as well?

Otherwise 16x14 and 14x14 could be fine as well :)

Is it usual to have (not have) a 1px padding around for the 16x16 ones?

dirtydancing:
I am not aware of a scheduled battery downsizing :)

I think that 16x14 is sweet, as said, I had not checked that size so far. It matches the battery icon size, and it matches the real-life look of an ethernet port :)

dirtydancing:
16x16 seems a bit over the top :)

palob:
Alright, what is more 16x14 seems the most squarish due to illusion.

wa4557:
Yeah let's make it 16x14; looks best

Foggalong:
+1 16x14 is sweet

dirtydancing:
OK, 16x14 is pushed, nice enhancement, ready to go :)

wired_previous
wired_proposed

wired_proposed_2

PR #483, opened 2015-05-04
Enhanced audio symbolic icons

dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/pull/482, this addresses https://github.com/numixproject/numix-icon-theme/issues/459.

The proposed audio symbolic icons are somewhat larger and thus more consistent with the other icons displayed in GNOME Shell. The alignment to grid is improved as well, all while keeping consistent with the current sound status icon design.

@wa4557 @palob As you have been part of the redesign of the corresponding wired icon: if you fancy to have a look, any thoughts on this one?

current
inkscape_current

proposed
inkscape_proposed

current volume icons
current_0
current_1
current_2
current_3

proposed volume icons
proposed_0
proposed_1
proposed_2
proposed_3

comparison of Adwaita, GNOME, and proposed
2_adwaita
2_gnome
proposed_2

comparison of previous, current and proposed
2_previous
current_2
proposed_2

In addition, also `audio-speakers-symbolic.svg` in `scalable/devices` needs to be adapted. In Numix base, this is a copy of `audio-volume-high-symbolic.svg` in `scalable/status`:

current
speakers_current

proposed
speakers_proposed


dirtydancing:
OK, I have done some more tweaking, and by now I think this is ready to be merged. I have cleaned up the circular alignment and straightened out the lowest bar, so that the icon is displayed less blurry.

Going from this one (cf. above)
proposed1

to this one
proposed2

proposed new
proposed2_0
proposed2_1
proposed2_2
proposed2_3

comparison current and proposed new
current_2
proposed2_2

For `audio-speakers-symbolic.svg` I have now used a symlink with the target `audio-volume-high-symbolic.svg`.

speakers2

palob:
The new one is better for sure. Maybe even a tad bigger?
How about a speaker without the curvature at the back?
Or without the spike at the front? Possibly better in combination with the new waves.

dirtydancing:
Indeed, I checked a couple of variants and came up with this one. This fits the new waves better, and it fits better with the other Numix GNOME Shell panel icons as well. The only reason why I did not get into this kind of design was to try and keep consistent with the fixed size sound icons. But I figure there is not really a reason for that, as GNOME Shell pretty much has its own set of Numix icons anyways.

Already pushed this, as this is convincing to me. If this goes on, I will have to switch to GNOME Shell :)

proposed3
proposed3_inkscape

proposed3 overview
proposed3_0
proposed3_1
proposed3_2
proposed3_3

comparison current proposed3
current_2
proposed3_2

proposed3 speakers
proposed3

palob:
Well done. A considerable improvement again :+1:

dirtydancing:
Thanks! :)

Foggalong:
If this were to be merged the change would have to be implemented across all sizes and platforms. I'm not sure on the new design either: it's leaning toward GNOME a little too much. That sort of thing would mean rethinking a lot more than just the volume icon too

For now, just the [original design you proposed](https://cloud.githubusercontent.com/assets/7164227/7461575/7e97da8a-f2ab-11e4-998c-8aa6197b38c9.png) would be sufficient to get this problem resolved.

palob:
Ahh yes, the last one looks pretty similar to GNOME/Adwaita.

dirtydancing:
While I think that even the last version is sufficiently distinct from GNOME/Adwaita, visually it indeed moves into that direction.

Comparison GNOME:
gnome

and latest version proposed and pushed above:
latest

Anyways, I have now gone back to the first version proposed above and ever so slightly modified it to be more "aligned to circle".

First proposed version from above:
proposed_1
 
Now proposed version:
proposed_aligned

which corresponds to this "alignment template" that I made use of only after the first version above:
proposed_aligned_grid

This slight clean-up results e.g. in a better distinction between the first and the second bar, which has bothered me a little with the first version proposed above.

Comparison first version proposed above and now proposed version below:
proposed_screen_1
proposed_screen_aligned

Comparison current version and now proposed version below:
current
proposed_screen_aligned

New proposed version:






speakers

Now proposed version is pushed and ready to be merged :)

PR #489, opened 2015-05-16
Adds dynamic calendar icons for all sizes

wa4557:
In Numix-Circle there's a icon `calendar.svg` in `48/apps` that is non circle and does not belong there. Morever since I created a PR in circle that allows to dynamically change the calendar icon [#2111] (https://github.com/numixproject/numix-icon-theme-circle/pull/2111))I created calendar icons for each day and each size. The icon is used for example in the ubuntu `indicator-datetime` and should be a base icon that works even if circle is not installed. 

I know that not all symbols are aligned to grid but I did my best. Unfortunately it's not possible to do 288 icons by hand :)


Foggalong:
If we're doing this it'd have to be as an addon to the Numix Circle theme rather than contained within it. For that reason I've made a [specific repo for the job](https://github.com/numixproject/numix-calendar-icons) :)

Foggalong:
All fine for me, but my only question would be why `/places` as the location? I would have thought `/apps` was the logical choice (plus has the benefit of being relatively empty)

wa4557:
Yeah I was thinking about this as well, but if there will be some kind of folder symlink from base `apps` to circle `apps` to fix the infamous status icon problem I figure that this will eventually leads to problems. Plus calendar isn't even a 100% app icon imo

Foggalong:
I was forgetting that. Places probably isn't the best place to put them, but it'll do until futher notice.

PR #488, opened 2015-05-15
Added usb alternative icon

dirtydancing:
`devices` directories up to size `48x48`. Addresses https://github.com/numixproject/numix-icon-theme/issues/101, cf. comments over there.

PR #497, opened 2015-05-23
Fixes #101

### LINKED ISSUE #101: Aren't these icons inverted?
Hello, guys. What's up?

Today, I noticed something that might be a little bug, I don't know. Take a look at the screenshot, please:

drives

As you can see in the picture, the first icon shows a red flash drive (if I'm not wrong), but it is actually my USB external 500 GB HDD; and the last icon shows a kind of a hard drive, but it is actually my USB SanDisk 8 GB flash drive. So... aren't those icons inverted?

I hadn't noticed this before because, since the day I started using Numix Icon Theme Circle, I've been using my flash drive only, and not my external HDD. And I thought that HDD icon would go with all removable media, so I didn't care. Then, today, when I plugged in my HDD, I noticed that red flash drive icon.

So, is that a bug?

Thanks for your time, guys. :)

brunces

wa4557:
I took up where @dirtydancing stopped. I wasn't sure as well which icons to turn into the harddisk. I did it to the best of my knowledge. Plus I can't test it because I don't know when these icons are used exactly so I suppose it's some kind of waiting for issue reports.


Foggalong:
@dirtydancing do you have the resources to check whether this pull will do the trick?

wa4557:
I think that nobody really knows where those icons are really needed. I tested using a USB-stick and an usb-harddrive and it worked (in ubuntu/unity), but those are just two of those icons I fixed. I have no idea if and by which distro the other icons are used

dirtydancing:
I checked with Arch/GNOME Shell 3.16.2 with USB stick and with USB harddrive (encrypted USB harddrive, so in addition to the icon a lock is shown for closed/open, working just fine): as expected, with current Numix base the USB stick icon is showing for both, and with this present pull request for the USB stick the USB stick symbol is showing, and for the USB harddrive the USB harddrive symbol.

This is parallel to my previous testing with Ubuntu 14.04 LTS earlier, cf. https://github.com/numixproject/numix-icon-theme/issues/101#issuecomment-58883307. The two relevant icon names here are

`drive-removable-media-usb.svg` and `drive-harddisk-usb.svg`

Just as @wa4557 I do not know how to test all of the icon names concerned here and listed over at https://github.com/numixproject/numix-icon-theme/issues/101#issuecomment-101771254.

Arguably, as also stated over there, it might make sense to keep the USB stick symbol only for `drive-removable-media-usb.svg` and use the USB harddrive symbol for `drive-harddisk-usb.svg` and for all the other icon names. In the sizes up to 48x48, all of these other icons are symlinks to `drive-harddisk-usb.svg`, so all of these icons would change to the USB harddrive symbol. I checked with Humanity icon theme some time ago while I was still using Ubuntu, and iirc also with that encompassing icon theme the only icon with the USB stick symbol is `drive-removable-media-usb.svg`.

On the other hand, I think that the selection chosen by @wa4557 makes sense as well and is apparently based on the icon names (possibly parallel to the sizes larger than 48x48 in Numix base?). Probably either way, this would mean to change it and wait for possible issue reports.

But I agree that a change should take place, because the distinction between the USB stick and the USB harddrive is a useful enhancement.

I am not sure about the sizes larger than 48x48, as there are currently no symlinks used there. This might be an opportunity to streamline all the fixed sizes to the same standard. But then again, the icons in the larger sizes are monochrome, so there seems to be no way to introduce a "red" harddrive for USB harddrive in those larger sizes.

As far as I can tell, for the larger sizes @wa4557 opted to symlink `drive-harddisk-usb.svg` to the target `drive-harddisk.svg`. This means a distinction from the USB stick symbol, but on the other hand there would no distinction to a non-USB harddrive (again, only for the sizes larger than 48x48).

dirtydancing:
Cf. also the conversation over at https://github.com/numixproject/numix-icon-theme/pull/488.

Issue #315, opened 2015-03-22
gtk-icons missing


hlechner:
![image](https://cloud.githubusercontent.com/assets/11244067/6769951/2f3c1dea-d08f-11e4-8c8d-06f878a715e5.png)
Left to Right:

**1º line**:
- [x] gtk-color-picker
- [x] gtk-connect
- [x] gtk-convert #284 #300 
- [x] gtk-dialog-authentication #285 
- [x] gtk-disconnect
- [ ] gtk-dnd
- [ ] gtk-dnd-multiple

**2º line**:
- [x] gtk-edit [#1309](https://github.com/numixproject/numix-icon-theme-circle/issues/1309)
- [x] gtk-select-color #284
- [x] gtk-select-font
- [x] gtk-index
- [x] gtk-undelete
- [x] gtk-orientation-landscape
- [x] gtk-orientation-portrait

**3º line**:
- [x] gtk-orientation-reverse-landscape
- [x] gtk-orientation-reverse-portrait
- [x] gtk-page-setup
- [x] gtk-print-error
- [x] gtk-print-paused
- [x] gtk-print-report
- [x] gtk-print-warning

Second gtk-dialog:
![image](https://cloud.githubusercontent.com/assets/11244067/6882493/7e685142-d565-11e4-8436-fffe9565f42c.png)
The second one is missing, that's why only the first appears.

- [x] gtk-caps-lock-warning
- [x] gtk-network

hlechner:
Its possible also to run the script that I ran on screenshot, you just need the gtk-dialog:

```
#! /bin/sh

pmap(){
echo '<pixmap icon_size="6" tooltip-text="'$1'"><input file stock="'$1'"></input></pixmap>'
}
export MAIN_DIALOG='
<vbox><hbox>
'"$(pmap gtk-color-picker)"'
'"$(pmap gtk-connect)"'
'"$(pmap gtk-convert)"'
'"$(pmap gtk-dialog-authentication)"'
'"$(pmap gtk-disconnect)"'
'"$(pmap gtk-dnd)"'
'"$(pmap gtk-dnd-multiple)"'
</hbox><hbox>
'"$(pmap gtk-edit)"'
'"$(pmap gtk-select-color)"'
'"$(pmap gtk-select-font)"'
'"$(pmap gtk-index)"'
'"$(pmap gtk-undelete)"'
'"$(pmap gtk-orientation-landscape)"'
'"$(pmap gtk-orientation-portrait)"'
[...]

PR #499, opened 2015-05-24
Makes dialog-password darker

wa4557:
The dialog-password icons in `*/status` were too light, so I made them darker. Two screenshots that show why they were too light:
Before
bildschirmfoto vom 2015-05-24 11 28 12

After
bildschirmfoto vom 2015-05-24 11 29 52


wa4557:
I added also symlinks to `gtk-dialog-authentication` since it was one of the icons @hlechner reported in #315

Issue #55, opened 2014-01-16
Network Conected Icon Missing


pauloelr:
The Network Conected Icon is Missing
![7939a098-7ee7-11e3-89e2-6dd2f284957b](https://f.cloud.github.com/assets/1790910/1934516/9a260ab8-7eeb-11e3-939e-656dc3e88d66.png)
Reporting as a new issue.
ping @me4oslav 
Thanks for being so polite.

PR #503, opened 2015-05-26
Deletes nm-device-wireless from 48x48/apps

wa4557:
Deletes `nm-device-wireless.svg` from `48/apps` since this icon is in `devices` and `notifications` and is no `apps` icon and nowhere used


Foggalong:
It was originally reported in #55 and then fixed in e5276d80455222a363445b5c8232bd0eb3c5e2d4 (pull #70). Is this not required anymore?

wa4557:
Nope. Because an appropriate (and arguably better) icon is now in `notifications` where it belongs. That's definitely no `apps` icon

Foggalong:
Ah, goodo!

PR #511, opened 2015-05-28
Adds gtk-convert

wa4557:
another one from #315


palob:
There's no 24x24 one. Is it needed?

Issue #124, opened 2014-05-20
Icon for virt-manager


clauded:
Some icons and buttons in virt-manager don't use (or have) Numix equivalent.

![capture du 2014-05-19 22 19 22](https://cloud.githubusercontent.com/assets/4673125/3026383/1f4f2f56-e014-11e3-97bb-b17554d8c511.png)
- [ ] 16x16/actions/icon_console.png
- [ ] 16x16/actions/vm_new.png
- [ ] 22x22/actions/icon_console.png
- [ ] 22x22/actions/icon_details.png
- [ ] 22x22/actions/vm_new.png
- [x] 22x22/devices/device_cpu.png
- [x] 22x22/devices/device_pci.png
- [x] 22x22/devices/device_usb.png
- [x] 22x22/devices/device_mem.png
- [ ] 22x22/devices/device_serial.png
- [ ] 24x24/actions/icon_console.png
- [ ] 24x24/actions/vm_delete_wizard.png
- [ ] 24x24/actions/vm_new.png
- [ ] 24x24/actions/vm_clone_wizard.png
- [ ] 24x24/actions/vm_import_wizard.png
- [ ] 24x24/actions/vm_new_wizard.png
- [ ] 32x32/actions/icon_console.png
- [ ] 32x32/actions/vm_new.png
- [ ] 32x32/status/state_paused.png
- [ ] 32x32/status/state_running.png
- [ ] 32x32/status/state_shutoff.png
- [ ] 48x48/actions/vm_clone_wizard.png
- [ ] 48x48/actions/vm_import_wizard.png
- [ ] 48x48/actions/vm_delete_wizard.png
- [ ] 48x48/actions/vm_new_wizard.png

bil-elmoussaoui:
Could you provide more information about this? like the icon lines of each icon, to help this get fixed asap.

gregorydk:
In /actions:
icon_console.png
icon_details.png
vm_clone_wizard.png
vm_delete_wizard.png
vm_import_wizard.png
vm_new.png
vm_new_wizard.png

In /devices:
device_cup.png
device_mem.png
device_pci.png
device_serial.png
device_usb.png

In /status:
state_paused.png
state_running.png
state_shutoff.png

bil-elmoussaoui:
I will symlink those that already exists : 
For devices:
device_usb.png to gnome-dev-flashdisk 
device_mem.png to gnome-dev-memory

gregorydk:
[...]

PR #512, opened 2015-05-29
changing the spotify-client name to spotify-indicator

bil-elmoussaoui:
Well , we now support spotify indicator icon in Hardcode-Tray fixer. But the icons that is used is 
spotify-panel (from the Numix-Circle in scalable) we decided to change the name to avoid this conflict


Foggalong:
Is `spotify-client.svg` not the icon that's used in the Status are in some cases?

wa4557:
Yeah, but it's also the name in circle, so this has to be changed anyway at some point. But yes symlinks are probably a better idea

bil-elmoussaoui:
My bad, i wrote ```spotify-panel.svg``` instead of ```spotify-client.svg```

wa4557:
@bil-elmoussaoui You're mixing some things here aren't you?

bil-elmoussaoui:
@wa4557 nop :p in the description i wrote " the icons that is used is spotify-panel " instead of spotify-client. and this icon is the same as the application icon and as this icon is in the ```scalable``` folder in Numix Circle this icon is used by Hardcode Tray instead of 22 or 24 size

wa4557:
That's not what I mean :) You're also symlinking something in devices in this Pull request :P

bil-elmoussaoui:
Ah :S the icons in devices are for #124 @Foggalong this can be merged like that o i have to separate them ?

Foggalong:
Separating them is needed. I don't have time to verify the pull request just yet :(

bil-elmoussaoui:
Okay :P i'm closing this then

PR #514, opened 2015-05-30
Inconsistency in gtk-color-chooser

wa4557:
One icon slipped through in one of my previous PRs


wa4557:
Edit: I pushed a complete overhaul, becaue the previous icons were bad made by me. These are much better

palob:
The new design is better indeed. Albeit I prefer the old brighter green hue (not worth reworking again for sure :) ).

Foggalong:
Like this :)

PR #522, opened 2015-05-31
some symlinks for virt-manager

bil-elmoussaoui:
For #124

### DIFF
--- a/Numix/128x128/devices/device_mem.svg
+++ b/Numix/128x128/devices/device_mem.svg
@@ -1,0 +1,1 @@
+gnome-dev-memory.svg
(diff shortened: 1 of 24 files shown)


wa4557:
You will have to do this in all sizes, I suppose

bil-elmoussaoui:
The application uses only those icons, so there's no need for bigger sizes

Foggalong:
It's a good idea to do it in ever size to remain flexible in case of upstream change and to make sure it work on LDPI, MDPI, HDPI, and XHDPI screens.

bil-elmoussaoui:
Done

wa4557:
256?

PR #524, opened 2015-05-31
Adds mail-inbox

wa4557:
Until now only outdated png icons were present for `mail-inbox`


wa4557:
Added mail-outbox as well

wa4557:
This fixes #100 btw

PR #525, opened 2015-06-03
Mimetype for Pascal source files

palob:
Modified Lazarus symbol (mascot of the Free Pascal Compiler is a cheetah).

Pushed the upper line:
allinone


wa4557:
I like the first one. Good work anyways, I always find it extremely hard to do mimetype icons :)

Issue #437, opened 2015-04-06
Elementary Ferya Support 


bil-elmoussaoui:
Hello,
I would like to report some missing icons/symlinks
1 - Geary support 
Using Numix Circle and Numix Base
![selection_001](https://cloud.githubusercontent.com/assets/7660997/7013879/c86e192a-dcbd-11e4-8ea0-460045705313.png)
Using elementary Theme
![selection_003](https://cloud.githubusercontent.com/assets/7660997/7013909/267c1d00-dcbe-11e4-8e37-2b5956989515.png)
- [ ] Archive : mail-archive : actions/16 and actions/24
- [ ] Tag : tag-new : actions/16,actions/24,actions/32,actions/48,actions/64,actions/128
- [ ] Flag : edit-flag and edit-mark : actions/16 and actions/24 for edit-flag ; actions/24 for edit-mark (could by symlinked as the icons are the same)
- [ ] Mail move : mail-move :  actions/24
That's all for this part of the application.
Without Numix:
![selection_004](https://cloud.githubusercontent.com/assets/7660997/7013987/1b378a50-dcbf-11e4-93e4-db2ff9835835.png)
With Numix 
![selection_005](https://cloud.githubusercontent.com/assets/7660997/7013991/33998f80-dcbf-11e4-94cb-cd6ff8ab6221.png)
Almost all the icons are supported only "Drafts" but unfortunately i didn't found the icon line..
2 - Shotwell 
With Numix : 
![selection_006](https://cloud.githubusercontent.com/assets/7660997/7014040/dfe89330-dcbf-11e4-96dd-95b3dcafe55b.png)
With Elementary 
![selection_007](https://cloud.githubusercontent.com/assets/7660997/7014052/07569930-dcc0-11e4-8a4e-4fcb3fa9c4ef.png)
- [ ] Image adjust : image-adjust : actions/24
- [ ] Image auto adjust : image-auto-adjust : actions/24
- [ ] Image red eye : image-red-eye : actions/24 

3 - Icons used in some apps : 
- [ ] Document import : document-import : actions/24 and actions/48

I will add more asap.
Thank you

wa4557:
Concerning your second screenshot: None of those icons are supported (except the trash-can) there are just a few .pngs in the Numix-folders that are remnants from the gnome-icon theme i suppose. So you can add them to your list. The "drafts" icon is probably `gtk-edit`

bil-elmoussaoui:
[...]

PR #532, opened 2015-06-06
symlink object-rotate-right to object-rotate

bil-elmoussaoui:
Fix one icon in #437

### DIFF
--- a/Numix/128x128/actions/object-rotate.svg
+++ b/Numix/128x128/actions/object-rotate.svg
@@ -1,0 +1,1 @@
+object-rorate-right.svg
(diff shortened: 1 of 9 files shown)


wa4557:
You should remove the symlink from 96x96. The icons in there are just the folder icons, no actual action icons

bil-elmoussaoui:
Done, thanks @wa4557

Foggalong:
@wa4557 @bil-elmoussaoui 

It looks as though GNOME is moving in the direction as using 96px as the default size for non-symbolic icons (they recently switched the default file manager size to it for example) so we may have to add them in future, but yeah for now it's not a problem

Issue #485, opened 2015-05-11
Small monochrome icons in Gnome shell


Foggalong:
Original brought to my attention in [this issue](https://github.com/moka-project/moka-icon-theme/issues/181#issuecomment-100892204) in the Moka icon theme.

> Starting with Gnome 3.16, Gnome shell shows small monochrome icons in the top panel. These Icons are available with the default applications of Gnome but other apps are missing these icons.
> Would you please consider adding these monochrome icons to your icons.

One that uses it:
![51d89ede-f74d-11e4-8eae-c150a1d06162](https://cloud.githubusercontent.com/assets/5920259/7565134/3d0152ec-f7e5-11e4-91ef-ef67233c34be.png)

One that doesn't:
![69054ae4-f74d-11e4-85ac-d3e1c5b99d18](https://cloud.githubusercontent.com/assets/5920259/7565212/1677db5e-f7e6-11e4-8374-4c01c4c116ba.png)

> If you provide `*-symbolic.svg` icons in `scalable/apps/` folder, Gnome shell will use those icons.

We should probably talk about what we want to do about this. Some points to discuss:
- Requires designing and adding ~1500 new icons
- Debatable whether they should be added here or in the app theme repos (i.e. Circle)
- Means having to test conflicts with the current `scalable/apps` symlinks which caters to HDPI screens
- Possible naming conflicts with some status icons

palob:
I think it's a bad design decision by the GNOME devs. It not only leads to symbolic icon/app icon inconcistencies  but also I think the Adwaita icons look dated or as if using the HighContrast theme.

I ended up disabling the Top Bar app icons altogether and possibly there will be an extension restoring the pre 3.16 behaviour.

It's a fairly limited case and I'd certainly prioritise other missing base theme icons but if someone feels like doing the icons go ahead :)

Thinking about a way to work around the symbolic icon display am I right in assuming that adding symlinks ``*-symbolic.svg`` for each and every file in ``scalable/apps`` would do the trick?

Foggalong:
Yup :/ we're basically given the choice now of either ditching GNOME from our inherits list or adding all the icons.

dirtydancing:
To me, this is a very questionable move of the GNOME devs as well.

Before further getting into this, one question: is this here to stay? I mean, why bother about it at all if this changes again with GNOME 3.18?

dirtydancing:
[...]

Issue #442, opened 2015-04-08
Icons missing for system settings in Ubuntu


dirtydancing:
Hi, in the course of the discussion in https://github.com/numixproject/numix-icon-theme/issues/438, the issue came up that apparently there are several system settings icons missing in Numix base.

This is a screenshot of daily Ubuntu 15.04 using Numix base (same in current Ubuntu 14.10):

![1504daily_numix](https://cloud.githubusercontent.com/assets/7164227/7049869/66e9526c-de1d-11e4-8787-8a5efb7b0278.png)

Not sure (yet) what the relevant icon names are, and whether this issue can be addressed without conflicts with the system settings icons in Numix-Circle (?). This is to open a separate issue report.

wa4557:
Also related to #160 is inheriting Humanity an alternative?

dirtydancing:
Yes, it should be. This issue relates to Ubuntu, and Ubuntu ships with Humanity.

One thing must be considered, though: The **differing inheritance lines**:

Currently in Ubuntu 14.10, Humanity inherits `gnome,hicolor` (which is in line with Numix base inheriting `gnome,hicolor`). But in Ubuntu 15.04, Humanity inherits `Adwaita,hicolor`, i.e. **Adwaita** instead of gnome. And because I figure that the proposed inheritance line of Numix would be `Humanity,gnome,hicolor`, there would be a deviation from that inheritance line in Humanity. This might not be a problem at all (after all, Adwaita is the successor of gnome), I just wanted to point that out.

Adwaita icon theme in current daily Ubuntu 15.04 does not have a specific inheritance line, and gnome icon theme does not either.

One thought that springs to mind with this, is that Numix could change its inheritance line to `Humanity,Adwaita,hicolor` as well, but that does not seem to be a good idea, because e.g. Ubuntu 14.10 does not ship with Adwaita icon theme (but Ubuntu 15.04 still also ships with gnome icon theme).

As an additional "bonus", inheriting from Humanity would also ensure that there are fallback icons for the Ubuntu notification icons, just in case. For Ubuntu 15.04, the notification icons have been moved from gnome to Humanity, cf. https://github.com/numixproject/numix-icon-theme/pull/298#issuecomment-83983789.

What do you make of this?

wa4557:
What happens if you are on KDE and Humanity is not present?

dirtydancing:
Humanity should simply be skipped if it is not present.

[...]

PR #537, opened 2015-06-07
adds Humanity to the inheritence instead of gnome, hicolor

### LINKED ISSUE #160: Missing additional drivers icon
There's no icon for additional drivers (jockey) on Trusty for numix and numix circle. Numix-shine and Numix-utouch have icons though.

Numix:
screenshot from 2014-08-05 21 51 55

Default:
screenshot from 2014-08-05 21 52 33

bil-elmoussaoui:
This will fix #160 as ```gnome, hiclolor``` doesn’t have any application icon, and ```Humanity``` inherits from ```gnome,hicolor```

### DIFF
--- a/Numix/index.theme
+++ b/Numix/index.theme
@@ -3,7 +3,7 @@ Name=Numix
 
 Comment=Icon theme from Numix Project
 
-Inherits=gnome,hicolor
+Inherits=Humanity
 
 Example=folder
 


dirtydancing:
Please cf. this conversation between @wa4557 and myself over at https://github.com/numixproject/numix-icon-theme/issues/442 about why `Humanity` should not be added to the inheritance line of Numix base.

And replacing `gnome,hicolor` with `Humanity` (and not keeping `gnome,hicolor`) leads to even more problems: what about distros that do not ship with `Humanity` at all?

hlechner:
Just to remember: I don't recommend it for non-ubuntu. gnome and hicolor came with gnome, but Humanity not, it's ubuntu theme.

dirtydancing:
@hlechner We posted at the same time, and obviously we agree on this :)

bil-elmoussaoui:
Adding Humanity to Gnome,Hicolor won't fix the problem for non-based on ubuntu distros neither?

wa4557:
No. If you're on KDE for example you would have to replace Humanity with oxygen. So your proposed fix is a very specific fix for Ubuntu based distros

bil-elmoussaoui:
That is not the best way to fix that :/ That will make it a long list of Inherits..

dirtydancing:
And this fix does not even fix all of Ubuntu itself: e.g. in Ubuntu 15.04, `Humanity` inherits from `Adwaita,hicolor` and not from `gnome` anymore. So how about older Ubuntu versions, esp. 14.04 LTS. Do these include `Adwaita`?

bil-elmoussaoui:
I'm using Ubuntu 14.04 and i have ```Adwaita``` pre-installed

dirtydancing:
Ah, OK. Anyways, as already stated in https://github.com/numixproject/numix-icon-theme/issues/442, inheritance lines of icon themes like e.g. `Humanity` might change and actually do change. This can cause further problems. imo, the current inheritance of Numix base from `gnome,hicolor` is the most workable for the time being.

bil-elmoussaoui:
I'm closing this then

wa4557:
Nevertheless, it might be a good idea to add Humanity and oxygen etc... (and other distribution gnostic themes) to the inheritence line without removing gnome/hicolor. After all they don't do any harm and it's basically the same with `gnome`. On KDE distros `gnome` isn't installed either, and Numix (with gnome in the inheritance line) still can be used on those distros...

palob:
@wa4557 Indeed, ``gnome`` might not be the universally best fallback. Isn't it even possible to set the inheritance depending on the desktop environment used?
And isn't the gnome icon set deprecated by Adwaita?

wa4557:
@palob Yes I think `gnome` will be / is replaced by `Adwaita`. If I were in charge I'd just add all those icon themes to the inheritance line (`Humanity`, `Adwaita`, `elementary`, `gnome`,`oxygen`,`hicolor` probably in this order), if the icon theme is not installed, well then the icons are not from those themes...

dirtydancing:
From what I have seen so far, the inheritances are tricky and complex. By adding further icon themes to the inheritance lines, complexity is added. And if complexity is added, there should be a clear advantage.

So far, e.g. for Humanity we have seen the `jockey` icon, which could also be added to Numix base, cf. https://github.com/numixproject/numix-icon-theme/issues/160. In addition, there is also e.g. the system settings overview in Ubuntu, cf. https://github.com/numixproject/numix-icon-theme/issues/442.

And given all the different distros/DEs/versions, everyone might be better served if users add inheritances according to their individual settings, rather than adding inheritances for all users by default.

In addition, I was under the impression that Numix rather aims to be independent of other icon themes such as gnome, cf. e.g. the conversation in https://github.com/numixproject/numix-icon-theme/issues/485. Adding more inheritances is the opposite direction.

Just raising some arguments here for the "status quo", I will be happy no matter how this turns out :)

Foggalong:
That aim is to eventually make Numix not need an inheritance line at all. If we can contain all of the icons from all of core themes mentioned above (and then keep up to date with them as they add new ones) then this is perfectly possible :)

dirtydancing:
From what I understand so far, the inheritance line in Numix could already be removed right now, at least when it comes to GNOME desktops. Please correct me if I am wrong, to me this is not entirely clear yet.

Let me explain:

Apparently, there are two "layers" of icon theme inheritances:
1) those specified by the icon theme itself and
2) those specified by the implementation of the freedesktop.org icon theme specification, http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html

For "Inherits" the specification states:

"(...) If no theme is specified implementations are required to add the "hicolor" theme to the inheritance tree. An implementation may optionally add other default themes in between the last specified theme and the hicolor theme."

And further down in the "icon lookup" the specification states:

"(...) The lookup is done first in the current theme, and then recursively in each of the current theme's parents, and finally in the default theme called "hicolor" (implementations may add more default themes before "hicolor", but "hicolor" must be last)."

It seems that the standard inheritances for GNOME encompass not only hicolor, but also "default" (in Arch/GNOME 3.16, the "default" icon theme is just an inheritance of Adwaita), so this would mean s.th. like `Inherits=default,hicolor`. So far I have not come across a current documentation of this, or for other implementations. Are you aware of such an implementation overview?

Anyways, this would mean that even if there is no inheritance line at all in Numix, in this case still "default" and hicolor would be inherited. This would also be consistent with the observation over at https://github.com/numixproject/numix-icon-theme/issues/485 that deleting `gnome` or the entire Numix inheritance line had no effect on the `*-symbolic.svg` icons being displayed as active app icons in the top bar of GNOME Shell 3.16.

Furthermore, this would also mean that adding more inheritances to Numix might actually cause more damage, because that could interfere with the "balance" of these inheritance implementations.

What do you make of this?

Foggalong:
That sounds like a great safefail system, but I'd sooner be safe in knowing we have all the icons covered too

PR #540, opened 2015-06-07
Adds gtk-dnd

wa4557:
`gtk-dnd` from #315. Might be too subtle but this was the only icon I was able to come up with...


Foggalong:
I think that design would be fine :) Maybe for dnd-multiple (feel free to include in this merge) just include a 3rd, slightly different coloured document inbetween the two

wa4557:
Done. Not as bad as I expected :) #315 can be closed now \o/

Issue #300, opened 2015-03-19
icon request: full gvim menu


chudur-budur:
These icons are missing (what I mean they come from the default gvim menu, not from numix), I was wondering is it possible to add sometime in near future ??

![gvim-other-1](https://cloud.githubusercontent.com/assets/1117795/6741123/87a855c4-ce5b-11e4-9b87-1f6d1e8d0167.png)
![gvim-other-2](https://cloud.githubusercontent.com/assets/1117795/6741126/87a99ac4-ce5b-11e4-9f1f-34cbefff68d3.png)
![gvim-other-3](https://cloud.githubusercontent.com/assets/1117795/6741125/87a97382-ce5b-11e4-80ba-a57aaeb0768c.png)
![gvim-save-all](https://cloud.githubusercontent.com/assets/1117795/6741124/87a95e2e-ce5b-11e4-8c55-261266bde9da.png)

thanks for your great work numix team !!!

Foggalong:
Do you have the icon names for those?

chudur-budur:
I could find only the last one, 

![gtk-save-all](https://cloud.githubusercontent.com/assets/1117795/6761336/532bc88a-cf28-11e4-9f59-372d5c3c5b91.png)

it's called gtk-save-all.png or gtk-save-all.svg

others I am trying to find out, let you know as soon as I get them.

Foggalong:
Thanks! Yeah, base icons are trickier to find

ChuckDaniels87:
Relate with #284 (at least gtk icons). We should make a list of these general gtk icons.

ChuckDaniels87:
A wiki entry for example?

hlechner:
Based on a gtk-dialog with gtk-icons on it, I filtered it and here the gtk icons that still missing:
![image](https://cloud.githubusercontent.com/assets/11244067/6769922/0dc5ca7c-d08e-11e4-916c-eefeff04f757.png)

If it's useful, I can create a new issue with the name of all of them

ChuckDaniels87:
It would be great, :wink:.

hlechner:
I've raised at #315 .

There are 21 icon missing in that list, but I'm not sure where I can find the complete list of gtk-icons

Foggalong:
Now contained in #315

PR #549, opened 2015-06-09
Added gtk-save-all.svg

palob:
Added ``gtk-save-all.svg``, missing according to #300.

Added missing ``.svg`` to ``geany-close-all``.


wa4557:
I think it's ok to add `gtk-save-all.svg`. I think however that these icons are not themeble in gvim and `gtk-save-all` is probably not used by gvim. They icons don't change if you use for example `hicolor` as icon theme nor any other icon theme

PR #575, opened 2015-06-13
Removes contact editor from base

wa4557:
Removes contact-editor.png from apps folders in base, because the icons are already present in circle, and are thus unnecessary in base


Foggalong:
Check yo diff

Issue #585, opened 2015-06-17
Icon for drive-harddisk-solidstate-symbolic.svg


dirtydancing:
In continuation of https://github.com/numixproject/numix-icon-theme/issues/583, Adwaita also contains `drive-harddisk-solidstate-symbolic.svg`. It has a distinct icon from `drive-harddisk-symbolic.svg`. This (untested) icon is missing from Numix base.

Issue #577, opened 2015-06-14
Missing icons from Adwaita theme


andia89:
Since there is the same issue report for Elementary freya #437 I figured I might as well file a bug report for all mising icons in the Adwaita theme (the new default icon theme for gtk 3+). There are actually not that much icons. 
- [x] network-wired-acquiring-symbolic
- [x] battery-symbolic
- [x] drive-harddisk-ieee1394-symbolic
- [x] drive-harddisk-system-symbolic
- [x] drive-harddisk-usb-symbolic
- [x] drive-harddisk-solidstate-symbolic
- [x] address-book-new-symbolic
- [x] airplane-mode-symbolic
- [x] alarm-symbolic
- [x] application-certificate-symbolic
- [x] application-exit-symbolic
- [x] application-rss+xml-symbolic
- [x] applications-engineering-symbolic
- [x] applications-games-symbolic
- [x] applications-graphics-symbolic
- [x] applications-multimedia-symbolic
- [x] applications-science-symbolic
- [x] applications-system-symbolic
- [x] applications-utilities-symbolic
- [x] application-x-addon
- [ ] application-x-appliance-symbolic
- [ ] application-x-executable-symbolic
- [x] application-x-firmware
- [x] appointment-missed
- [x] appointment-missed-symbolic
- [x] appointment-new-symbolic
- [x] appointment-soon
- [x] appointment-soon-symbolic
- [x] audio-card-symbolic
- [x] call-missed-symbolic
- [x] call-stop-symbolic
- [ ] camera-video
- [x] camera-video-symbolic
- [x] changes-allow
- [x] changes-allow-symbolic
- [x] changes-prevent
- [x] channel-insecure-symbolic
- [x] channel-secure-symbolic
- [x] checkbox-checked-symbolic
- [x] checkbox-mixed-symbolic
- [x] checkbox-symbolic
- [x] color-select-symbolic
- [ ] colorimeter-colorhug-symbolic
- [x] computer-apple-ipad-symbolic
- [ ] computer-fail
- [x] computer-fail-symbolic
- [x] computer-symbolic
- [x] display-projector-symbolic
- [x] document-print-preview-symbolic
- [x] document-print-symbolic
- [x] document-revert-rtl
- [x] document-revert-symbolic
- [x] document-revert-symbolic-rtl
- [ ] drive-multidisk
- [ ] drive-removable-media-symbolic
- [x] edit-clear-all
- [x] edit-clear-rtl
[...]

PR #586, opened 2015-06-17
Added drive-harddisk-solidstate-symbolic.svg

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme/issues/585.

drive-harddisk-solidstate-symbolic

The design is based on and in line with `drive-harddisk-symbolic.svg`:

drive-harddisk-symbolic

### DIFF
--- a/Numix-Light/scalable/devices/drive-harddisk-solidstate-symbolic.svg
+++ b/Numix-Light/scalable/devices/drive-harddisk-solidstate-symbolic.svg
@@ -1,0 +1,66 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   viewBox="0 0 16.0002 16"
+   style="enable-background:new"
+   id="svg2"
+   version="1.1"
+   inkscape:version="0.91 r13725"
+   width="100%"
+   height="100%"
+   sodipodi:docname="drive-harddisk-solidstate-symbolic.svg">
+  <metadata
+     id="metadata14">
+    <rdf:RDF>
+      <cc:Work
+         rdf:about="">
+        <dc:format>image/svg+xml</dc:format>
+        <dc:type
+           rdf:resource="http://purl.org/dc/dcmitype/StillImage" />
+        <dc:title></dc:title>
+      </cc:Work>
+    </rdf:RDF>
+  </metadata>
+  <defs
+     id="defs12" />
+  <sodipodi:namedview
+     pagecolor="#ffffff"
+     bordercolor="#666666"
+     borderopacity="1"
+     objecttolerance="10"
+     gridtolerance="10"
+     guidetolerance="10"
+     inkscape:pageopacity="0"
+     inkscape:pageshadow="2"
+     inkscape:window-width="1280"
+     inkscape:window-height="773"
+     id="namedview10"
+     showgrid="true"
+     inkscape:zoom="34.0625"
+     inkscape:cx="8.0001"
+     inkscape:cy="8"
+     inkscape:window-x="0"
+     inkscape:window-y="27"
+     inkscape:window-maximized="1"
+     inkscape:current-layer="svg2">
+    <inkscape:grid
+       type="xygrid"
+       id="grid3833" />
+  </sodipodi:namedview>
+  <g
+     id="g3759"
+     style="fill:#bebebe;fill-opacity:1">
+    <path
+       sodipodi:nodetypes="ssssssssssssssss"
+       inkscape:connector-curvature="0"
+       id="rect2984"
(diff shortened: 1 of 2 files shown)


pgera:
This shows some weird behaviour, which is possibly outside the scope of the icon theme. It incorrectly shows the linux partition of my SSD as a regular hard drive. Oddly, it correctly identifies the ntfs partitions of the same hard drive as solid state.

dirtydancing:
Thanks for the feedback, great to have someone with a SSD drive around :)

1) Let's see: my guess is that the linux partition of your SSD is the "system harddisk". If possible, could you check by changing `drive-harddisk-system-symbolic.svg` if that changed icon is displayed.

iirc when I checked in Nautilus with Arch/GNOME (3.16), even for the system harddisk/partition, the icon `drive-harddisk-symbolic.svg` was used (regular HDD symbol), not `drive-harddisk-system-symbolic.svg` (not present in Numix base before, and now a symlink with the target `drive-harddisk-symbolic.svg`, just like in default Adwaita).

I will check again later as well.

2) This leads to my second question: is this handled differently with default Adwaita icon theme?

pgera:
Yes, it is the "system harddisk" partition. However, it is not getting the icon from drive-harddisk-system-symbolic.svg. It is actually getting it from drive-harddisk-symbolic.svg. Changing the latter changes the icon. Also, the same issue affects other icon themes (humanity. faba-mono, hicolor, moka). Adwaita doesn't distinguish b/w hdd and ssd. I still think this is possibly related to lvm, since that's an additional layer on top of the hard drive. For eg., it is also possible to combine multiple disks (of possibly different types) into one volume, in which case you would have to choose to show it one way or the other, and they probably chose hdd.

dirtydancing:
@pgera OK, so this confirms that `drive-harddisk-symbolic.svg` and not `drive-harddisk-system-symbolic.svg`  is used also for the "system harddisk" (at least in a LVM setup).

This might be LVM-related indeed. I also use LVM (on LUKS), but in my setup only one harddisk/partition is displayed in Nautilus, so I cannot really test for more complex setups.

You are writing that Adwaita does not distinguish between HDD and SSD. This might be related to the version of Adwaita. Apparently, `drive-harddisk-solidstate-symbolic.svg` is a more recent addition, cf. this comment https://github.com/numixproject/numix-icon-theme/issues/577#issuecomment-112715468.

@Foggalong Unless further issues relating to this come up, I would suggest to leave everything as it is.

dirtydancing:
@wa4557 @palob Sorry, I originally did not want to bring you in once again, but if you fancy to have a look at this thread and check what kind of `drive-harddisk-*-symbolic` icons are displayed in your setups (the symbolic icons e.g. in Nautilus), that might actually shed some more light on this situation.

wa4557:
it's `drive-harddisk-symbolic.svg` in Ubuntu using a ssd as system harddisk

dirtydancing:
@wa4557 Is this a LVM setup?

wa4557:
@dirtydancing No

dirtydancing:
Well, after this, I believe it is save to conclude that `disk-harddrive-symbolic.svg` is always used for the "system harddisk". This means that `disk-harddrive-system-symbolic.svg` currently is merely a place holder. And this means that there is currently no ootb way that a system harddisk/partition SSD will be displayed with `disk-harddrive-solidstate-symbolic.svg`, as it would not make sense to use the SSD symbol for the general `disk-harddrive-symbolic.svg`.

The only question remaining now imo is whether to change `disk-harddrive-solidstate-symbolic.svg` into a symlink with the target `disk-harddrive-symbolic.svg`. This might lessen any confusion in cases like above when a SSD is only in part used as "system harddisk", which currently means that the HDD symbol is displayed for the SSD system partition, and the SSD symbol for the other SSD partitions. But then again, there are also use cases in which the differentiation between HDD symbol and SSD symbol is fitting (e.g. an additional SSD as a non-system harddisk).

If there was an icon named `disk-harddrive-hdd-symbolic.svg` or the like, it would be possible to use a "general" icon for `disk-harddrive-symbolic.svg`, such as currently `disk-harddrive-symbolic-backup/backup2.svg` in Numix base. But that is currently not the case, so this is not an option, either. Those `-backup/backup2` icons would actually also make sense as "system harddisk" icons, but as we have seen, `drive-harddisk-system-symbolic.svg` is currently not picked up.

@Foggalong Summing up, my recommendation would still be to leave things as they currently are.

palob:
I had a look and the symbol for *"Computer"* (shortcut to ``/``) is always the HDD symbol.

When booting into a live iso with my "big laptop" the unmounted SSD partitions are displayed incosistently: ``disk-harddrive-solidstate-symbolic.svg`` for the boot partition, ``drive-removable-media-szstem-media-symbolic.svg`` (no Numix icon for this one) for all other partitions.

Booting into a live iso on my Chromebook however shows all unmounted SSD partitions as ``disk-harddrive-solidstate-symbolic.svg``.

The SSD is the single system drive on both systems, no LVM used.

dirtydancing:
Wow, this is getting more and more complex by the hour :)

So we have yet another inconsistency. This one is a very odd one: Where does `drive-removable-media-system(-media?)-symbolic.svg` come from?

This is not in current Numix base (only `drive-removable-media-usb-symbolic.svg`), and it is not on the list of currently misssing icons from Adwaita icon theme over at https://github.com/numixproject/numix-icon-theme/issues/577 (only `drive-removable-media.svg`). Actually, Adwaita icon theme in Arch/GNOME only has `drive-removable-media-symbolic.svg`, not the more specific `drive-removable-media-usb-symbolic.svg`.

And there should at least be a distinction between "harddisks" (be they HDD, SSD, USB, FireWire, what have you) and "removable media" such as USB sticks. Why a previously never heard of "removable media" icon is displayed for SSD partitions is beyond me.

What kind of wonderland have we entered here? :) Oh, how sweet were the times when there was only `drive-harddisk-symbolic.svg` in Numix base, and no one seemed to have any issues. Given this new situation, I am becoming open-minded to change `drive-harddisk-solidstate-symbolic.svg` into a symlink pointing to `drive-harddisk-symbolic.svg`, so that one will be the catch-all for harddisks (except for the fringe cases like the scenario above).

In the end, this could make things as consistent as possible, with `drive-harddisk-symbolic.svg` on the one hand and `drive-removable-media-symbolic.svg` on the other.

What do you think? `drive-harddisk-solidstate-symbolic.svg` as yet another symlink with the target `drive-harddisk-symbolic.svg`?

palob:
Sorry , it's just ``drive-removable-media-media-symbolic.svg``. 
I'd simply keep the current state (with the PR merged) which provides improved SSD/HDD distinction.

It's just not possible to have a SSD icon for  "Computer" (shortcut to /), if that's what you are looking for.

dirtydancing:
Agreed that for the time being, leaving things as they currently are is the way to go.

Issue #580, opened 2015-06-16
Missing kde icons from base


wa4557:
Ok apart from Adwaita, Elementary and gnome oxygen also misses a lot of base icons (thankfully a lot of symlinks are possible in this one). Be warned the list is extremely long :)

- [ ] acrobat
- [ ] align-horizontal-bottom-out
- [ ] align-horizontal-center
- [ ] align-horizontal-left
- [ ] align-horizontal-left-out
- [ ] align-horizontal-right
- [ ] align-horizontal-right-out
- [ ] align-horizontal-top-out
- [ ] align-vertical-bottom
- [ ] align-vertical-bottom-out
- [ ] align-vertical-center
- [ ] align-vertical-top
- [ ] align-vertical-top-out
- [ ] application-relaxng
- [ ] applications-development-translation
- [ ] applications-development-web
- [ ] applications-education
- [ ] applications-education-language
- [ ] applications-education-mathematics
- [ ] applications-education-miscellaneous
- [ ] applications-education-preschool
- [ ] applications-education-school
- [ ] applications-education-science
- [ ] applications-education-university
- [ ] applications-toys
- [ ] application-sxw
- [ ] application-vnd-google-earth-kml
- [ ] application-x-abiword
- [ ] application-x-applix-word
- [ ] application-x-awk
- [ ] application-x-chm
- [ ] application-x-font-snf
- [ ] application-x-gzpostscript
- [ ] application-x-it87
- [ ] application-x-java-applet
- [ ] application-x-k3b
- [ ] application-x-kcsrc
- [ ] application-x-kexi-connectiondata
- [ ] application-x-kexiproject-shortcut
- [ ] application-x-kexiproject-sqlite
- [ ] application-x-kexiproject-sqlite2
- [ ] application-x-kexiproject-sqlite3
- [ ] application-x-kformula
- [ ] application-x-kgetlist
- [ ] application-x-kplato
- [ ] application-x-kvtml
- [ ] application-x-kword
- [ ] application-x-marble
- [ ] application-x-mimearchive
- [ ] application-x-mplayer2
- [ ] application-x-mswinurl
- [ ] application-x-mswrite
- [ ] application-x-nzb
- [ ] application-x-plasma
- [ ] application-x-qet-element
- [ ] application-x-qet-project
- [ ] application-x-sharedlib
- [ ] application-x-smb-server
- [ ] application-x-smb-workgroup
- [ ] application-x-srt
- [ ] application-x-srtrip
[...]

PR #587, opened 2015-06-18
QML mimetype icons

palob:
overview-qml


palob:
I've just seen it's one from #580

Issue #588, opened 2015-06-19
Icon for drive-removable-media-symbolic.svg


dirtydancing:
A Numix base icon is missing for `drive-removable-media-symbolic.svg`. This icon is present in current GNOME default Adwaita icon theme. But in contrast to Numix base, Adwaita does not contain `drive-removable-media-usb-symbolic.svg`.

So the way to go about this would be to copy the current Numix base `drive-removable-media-usb-symbolic.svg` and rename the copy into `drive-removable-media-symbolic.svg`. I checked with Arch/GNOME: in the absence of `drive-removable-media-usb-symbolic.svg`, then `drive-removable-media-symbolic.svg` is picked up for USB sticks.

It should be fine to take this icon as the "general" one for "drive-removable":

![drive-removable-media-symbolic](https://cloud.githubusercontent.com/assets/7164227/8252995/8a9cd53e-168b-11e5-8929-527a451335d2.png)

At the same time, it then makes sense to rename the current "backup" icons in Numix base, naming in line with `drive-harddisk-symbolic-backup...`, cf. https://github.com/numixproject/numix-icon-theme/issues/578:
`drive-removable-media-usb-symbolic-1.svg` to `drive-removable-media-usb-symbolic-backup.svg`
`drive-removable-media-usb-symbolic-old.svg` to `drive-removable-media-usb-symbolic-backup2.svg`
`drive-removable-media-usb-symbolic.svg` to `drive-removable-media-usb-symbolic-backup3.svg`

palob:
Closing this. Don't think we need our own symbolic icons. The Adwaita ones blend in well enough.

PR #589, opened 2015-06-19
Added drive-removable-media-symbolic.svg

dirtydancing:
Fixes https://github.com/numixproject/numix-icon-theme/issues/588.


palob:
Hmm, the Adwaita icon ``drive-removable-media-symbolic.svg`` looks more universal whereas yours shows a USB thumb drive.

As indicated in #586 however ``drive-removable-media-symbolic.svg`` isn't limited to USB thumb drives.

dirtydancing:
Well, we might take a more generic one as well. As this icon is not showing a specific symbol such as the USB symbol, it is already pretty universal imo. Another aspect to consider is that there is also `media-removable-symbolic.svg` in Adwaita, which in Adwaita is a USB thumb drive symbol, different to `drive-removable-media-symbolic.svg`.

I was thinking of adding `media-removable-symbolic.svg` in the next step as a symlink pointing to `drive-removable-media-symbolic.svg`. But again, we could also go for a different, even more universal icon here for `drive-removable-media-symbolic.svg`.

@palob iirc the case that you mentioned was an SSD harddrive incorrectly not picking up `drive-harddisk-solidstate-symbolic.svg`, but rather `drive-removable-media-symbolic.svg`?

dirtydancing:
To illustrate, this is `drive-removable-media-symbolic.svg` in Adwaita:

adwaita

This icon is looking somewhat similar to Numix base's `drive-harddisk-symbolic-solidstate-symbolic.svg`.

palob:
@dirtydancing Yes, that's how it was.

Just thought maybe this icon is used for external SSD or something.

dirtydancing:
@palob I see what you are saying. I am looking into it and seeing if I can come up with an alternative icon. The currently proposed "USB" thumb drive icon would also fit with the non-symbolic icon in the notification in GNOME Shell btw.

Anyways, it might even be worth considering to keep `drive-removable-media-usb-symbolic.svg` with the current "USB" thumb drive icon and add an additional, even more universal, icon for `drive-removable-media-symbolic.svg`. That usb icon is still picked up over the more general `drive-removable-media-symbolic.svg`, although `drive-removable-media-usb-symbolic.svg` is not in current Adwaita. On the other hand, this might cause more inconsistencies from case to case. Alas, quite some considerations again... :)

dirtydancing:
As to the external SSD drive that you mentioned: That would indeed be interesting to see. I can tell that e.g. an external USB HDD is displayed with `drive-harddisk-usb-symbolic.svg` or more generally with `drive-harddisk-symbolic.svg`. This seems to indicate that usually "harddisks" are represented by "harddisk" symbols. But then again, this seems to differ in detail, as far as we have seen.

dirtydancing:
How about this one for `drive-removable-media-symbolic.svg`?:

alternative

This is as universal as it gets, and it would blend in quite nicely with the already existing `drive-harddisk-symbolic.svg` and `drive-harddisk-solidstate-symbolic.svg`:

drive-harddisk-symbolic

drive-harddisk-solidstate-symbolic

The icon design is somewhat inspired by the corresponding Ozon icon (another great icon theme), but imo this would still be a distinct Numix icon. And btw, neither Adwaita nor Ozon are providing this kind of blending in with the other `harddisk-symbolic` icons :)

This is of advantage, because as we have seen, in detail these icons are used somewhat interchangeably, so an integrative fit like this would ensure as much visual consistency as possible.

Ozon icon:

ozon

palob:
:+1: 
If its still considered too Ozony maybe rotate the arrows through 90°.

dirtydancing:
Here we go, Alternative 2:

alternative_2

Foggalong:
What exactly is this icon for? We use the two arrow thing for network stuff so I'm not sure we should be using it here

palob:
Maybe just one arrow to symbolise removability?

dirtydancing:
This is for `drive-removable-media.svg`, which will be for USB thumb drives and arguably other "removable" drives. The arrows are signaling plugging it in an plugging it out (=removability). Also they are signaling interchanging data inbetween the two.

The "frame" of the Numix harddisk symbol distinguishes this one from network imo.

Foggalong:
Don't we have symbols already for USB sticks and harddrives?

palob:
This one's missing and GNOME shows the Adwaita one instead.

dirtydancing:
Haha, this is quite complex/complicated :) I cannot reproduce the entire conversation that has lead to this point right now, but how about Alternative 3?:

alternative_3

dirtydancing:
Of course, I am happy to continue this conversation later. Just this: as we are currently integrating the icons that default Adwaita is providing, this one is in Adwaita, but `drive-removable-media-usb-symbolic.svg` is not. So apparently this one is the new catch-all for removable "drives", whereby the borders to the `drive-harddisk-*-symbolic.svg` icons appear to be somewhat fluent.

palob:
The best draft so far imo. Perhaps better with a shorter (1 px) arrow to increase the distance to the horizontal line.

dirtydancing:
OK, I have checked a couple of variants for the more "universal" design for the removable media, and I came up with this one. What do you make of this, Alternative 4?:

alternative_4

Also works together nicely with the always present "Computer" `drive-harddisk-symbolic.svg`:

drive-harddisk-symbolic

as well as with `disk-harddrive-solidstate-symbolic.svg` of course:

drive-harddisk-solidstate-symbolic

dirtydancing:
Oh, and we could still make use of the current Numix "USB" thumb drive icon for `media-removable-symbolic.svg`, as explained above. But that would be a separate issue report/PR.

dirtydancing:
To illustrate, here is a Nautilus/Files 3.16.2 screenshot with the latest Alternative 4 icon. Displayed in the former "devices" category are "Computer", a 4 GB USB thumb drive and a 500 GB USB HDD ("Backup"):

numix

In comparison, this is what it looks like with current Adwaita:

adwaita

palob:
Makes sense to me.

dirtydancing:
Shall I push Alternative 4 then for `drive-removable-media-symbolic.svg`? The current thumb drive icon is kept as "-backup3" anyways, and will then also be the right icon for `media-removable-symbolic.svg`.

palob:
I'd say yes. If need be for a maximum confusion-proof design you could go without the arrows.

dirtydancing:
Interesting idea. So the same icon for `drive-removable-media-symbolic.svg` and for `drive-harddisk-solidstate-symbolic.svg`? Why not. Actually, to be ultimately confusion-proof, we might also just use the current `drive-harddisk-solidstate-symbolic.svg` for all these icons. Well, maybe too simple :)

dirtydancing:
But I would recommend as well to go for Alternative 4, as proposed above.

imo possible confusion will not come from these different icons, but rather from them not being used properly (like e.g. the internal SSD of @palob showing up as `drive-removable-media-symbolic.svg`).

wa4557:
I really like the last one :+1:

dirtydancing:
@Foggalong Any news on this issue? I still prefer Alternative 4 as well. Should I push that one?

bbf1cdf6-16db-11e5-80bd-cb86aea6758a

Foggalong:
I'll confess, I got a little lost on this one - in what context is this icon used?

palob:
I got it shown (erroneously?) for unmounted SSD partitions when booting a live iso. Maybe it's used for external SSDs?

dirtydancing:
No problem :) This is for `drive-removable-media-symbolic.svg`, the new "catch-all" for removable media e.g. in the left sidebar of Nautilus/Files. It supersedes the current `drive-removable-media-symbolic-usb.svg`. That USB stick icon will be kept in Numix base as a backup in any case with this PR.

Additionally, in a follow-up issue/PR, the current symbolic USB stick symbol will then be the icon for the new `media-removable-symbolic.svg`. I am not sure where exactly `media-removable-symbolic.svg` is used, but that one is in current default Adwaita icon theme. Again, that would be a separate issue, but the USB icon will not be "lost".

Foggalong:
I'm confused then, why are we replacing this icon?

dirtydancing:
@palob Yeah, external SSD's seem to be a kind of borderline case. There is `disk-harddisk-solidstate-symbolic.svg`, but that one seems to be used mainly for internal SSD. Then there is `drive-harddisk-symbolic.svg` which is used for "Computer", the system partition (apparently no matter if HDD or SSD), and which is also used e.g. for external USB HDD harddrives. In contrast, the icon in question here, `drive-removable-media-symbolic.svg`, is used e.g. for external USB sticks, and might also be used for external SSD drives (?).

dirtydancing:
@Foggalong We are proceeding according to the icons in Adwaita, and Adwaita does not contain `drive-removable-media-usb-symbolic.svg` anymore.

palob:
Since when? I think it's better to keep the old icons to retain compatibility to non-bleeding edge distros or distros/DEs which don't depend on Adwaita in the first place.
Is it a matter of whatsoever clashes?

Foggalong:
Yeah, we're definitely nor removing icons just based on what Adwaita uses.

dirtydancing:
Well, we are not "removing" the icon specifically for external USB sticks. These are just covered by `drive-removable-media-symbolic.svg`.

And of course, we could still keep `drive-removable-media-usb-symbolic.svg` as it presently is in Numix base. That one will still be displayed for USB sticks specifically, even if `drive-removable-media-symbolic.svg` is present.

As argued above, I still think it is better to remove/rename `drive-removable-media-usb-symbolic.svg`, because the distinction between these "drive" symbols is somewhat blurred, and it might actually be less confusing without `drive-removable-media-usb-symbolic.svg`.

Actually, this entire thread started because @palob made a valid point: `drive-removable-media-symbolic.svg` in Adwaita looks more universal than the USB stick symbol. And it is the intention to keep these "drive" icons as universal as possible. This is also why it makes sense to not keep the non-universal `drive-removable-media-usb-symbolic.svg`. That icon might also be (wrongly?) used e.g. for an external USB SSD harddrive (?), and that would then be confusing if a "USB" stick symbol is displayed.

dirtydancing:
I am not sure in which previous version of Adwaita, if at all, there was `drive-removable-media-usb-symbolic.svg`. This icon name does not exist in several other icon themes, either, e.g. Ozon.

palob:
But how to be sure ``drive-removable-media-symbolic.svg`` is the fallback for ``drive-removable-media-usb-symbolic.svg``?

dirtydancing:
Well, I figure we cannot be sure for all distros whether `drive-removable-media-symbolic.svg` is the fallback for `drive-removable-media-usb-symbolic.svg`.

On the other hand, I am not even sure where `drive-removable-media-usb-symbolic.svg` is coming from at all. E.g., I have checked the Adwaita (3.16/3.14) packages and the preceeding gnome symbolic packages (down to 3.4) in Ubuntu, and nowhere does `drive-removable-media-usb-symbolic.svg` appear. So far, I have also not come across it in previous releases over at https://github.com/GNOME/adwaita-icon-theme/releases. So, arguably this icon is "outdated" :)

Anyways, the fallback point is a valid one, and thus the most pragmatic way to go about this would arguably be to keep `drive-removable-media-usb-symbolic.svg`, just as a symlink with the target `drive-removable-media-symbolic.svg`, so as to ensure consistency. What do you make of this?

dirtydancing:
OK, I have pushed the variant that I believe is the most suitable after our latest conversation above.

dirtydancing:
To clarify: also this variant encompasses the three `*-backup` icons as laid out in the issue report https://github.com/numixproject/numix-icon-theme/issues/588.

dirtydancing:
@Foggalong I still think this is ready to be merged :) Comparative screenshots of `scalable/devices`:

**Current:** (relevant are the `drive-removable-media-*.svg` icons)

current

**Proposed:** (the added arrow stands for symlink)

proposed

Foggalong:
There are more base icon themes out there than just Adwaita for `drive-removable-media-usb-symbolic.svg` to come from. I'll be honest, I'm not sure I'm 100% comfortable with this solution either. Unlike previous changes like this we don't seem to know very much about what's used where or why.

dirtydancing:
Yeah, this is a somewhat blurry situation. Still, I would recommend to go for this change. Are you thinking of an alternative approach to this? Possibly add the suggested `drive-removable-media-symbolic.svg` and at the same time leave the current `drive-removable-media-usb-symbolic.svg` as it is?

Foggalong:
I'm still not sure one this. I can understand the logic behind the change but it seems very dependent on what's going on in the latest GNOME/Adwaita :/

dirtydancing:
Well, I could close this PR, and we could come back to this subject e.g. in case of future issue reports.

Foggalong:
Sure :+1:

dirtydancing:
Just a quick update: Gnome 3.18 has landed in Arch, and Files/Nautilus 3.18 now displays the removable media icons under "Other Locations" on the right hand side with non-symbolic icons.

Screenshot:

nautilus_318

wa4557:
@dirtydancing Is this good or bad?

dirtydancing:
Well, it makes this issue less relevant. But I am not exactly sure what is going on here, because a different USB pendrive was still showing with a symbolic icon on the left hand side in Files/Nautilus.

palob:
I've put the 3.18 update on hold and I use Nemo as my main file manager anyway but does that mean you have to go to the "Other Locations" view every time you want to do things like unmounting external devices?

dirtydancing:
I have only just updated to 3.18, and so far I am noticing that e.g. one USB pendrive was showing under "Other Locations", whereas another one was showing on the left hand side like before. Maybe there is a method to this :)

Issue #132, opened 2014-06-13
Nemo drive sidebar icon.


ghost:
Nemo and thunar drive sidebar icon is too small and looks fuzzy.

![Alt Text](http://i.imgur.com/3f9tGXx.png)

![Alt Text](http://i.imgur.com/SazI7cG.png)

Also the nemo drive eject button looks kinda weird compared to thunar eject button, could you fix it?

This is the icon nemo drive eject button should use:
`/usr/share/icons/Numix/scalable/actions/media-eject-symbolic.svg`

I think this is the drive icon for nemo and thunar sidebar:
`/usr/share/icons/Numix/16x16/devices/drive-removable-media.svg`

PR #599, opened 2015-06-26
Added nemo-eject.svg

palob:
Added ``nemo-eject.svg``. It's the drive eject icon for Nemo.
Original design present in ``hicolor/16x16/actions`` and ``hicolor/32x32/actions``:
nemo-eject

New design:
nemo-eject

The new one however will be shown only when just the Numix base theme is chosen.
With Numix Circle selected for some reason still the app icon ``nemo.svg`` is shown (related to #132)

### DIFF
--- a/Numix/128x128/actions/nemo-eject.svg
+++ b/Numix/128x128/actions/nemo-eject.svg
@@ -1,0 +1,6 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
+ <g transform="matrix(6.6666667,0,0,8.0000001,10.666666,-1.04e-6)" fill="#dc322f" fill-opacity="1">
+  <path d="m 14 11 0 2 -12 0 0 -2 z"/>
+  <path d="m 2 9 6 -6 6 6"/>
+ </g>
+</svg>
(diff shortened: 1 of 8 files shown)


dirtydancing:
@palob Does `nemo-eject.svg` show correctly when it is placed in Numix-Circle? Because if it does, this might be yet another issue with the order of the icon lookup and the inheritances: it might be that `nemo-eject.svg` is only looked up if it is in the "first" icon theme in the line (or at least not "behind" `nemo.svg`), and that `nemo.svg` as a kind of fallback icon is picked up before moving further along the inheritance line (to Numix base, which Numix-Circle inherits from). This behaviour might be related to the infamous https://github.com/numixproject/numix-icon-theme/issues/190 (?).

dirtydancing:
Actually, I just quickly checked with Arch/GNOME: reproduced that when adding `nemo-eject.svg` to `Numix/16x16/actions`, that icon only shows when Numix base is selected (once again, an icon cache updated appears to be necessary), whereas in Numix-Circle still the Circle app icon `nemo.svg` is shown.

Tested: added `nemo-eject.svg` to `Numix-Circle/48x48/apps` (for simplicity's sake, because that directory already exists). Result: `nemo-eject.svg` is correctly displayed when Numix-Circle icon theme is selected. This indicates that the presence of `nemo.svg` does not "block" `nemo-eject.svg`, but that rather indeed there might be a problem with the icon lookup order and the inheritances.

dirtydancing:
Just to clarify: The (possible) details about icon lookups/inheritances should not keep `nemo-eject.svg` from being added to Numix base. This addition makes sense in any event.

If anything, this might turn out to be yet another piece in the puzzle that is https://github.com/numixproject/numix-icon-theme/issues/190, which has been causing e.g. the Numix-Circle PPA to be on standby for several months now.

PR #607, opened 2015-07-01
Missing indicator keyboard symlinks

bil-elmoussaoui:
Some more symlinks for #437


Foggalong:
* One merge per feature/type of icon that you're adding
* As I understand it the indicators can't just be linked together for different numbers, all of the number signify different layouts that need to be distinguished in their icons. That's one of the main reasons tackling the keyboard indicators is such an unpopular task

PR #608, opened 2015-07-01
adds three symlinks for freya

bil-elmoussaoui:


Foggalong:
Are you sure that `battery-ac-adapter.svg` isn't a status icon and that adding it is going to completely topple the delicately balanced mess that is the battery status icons.

bil-elmoussaoui:
My bad :P yes the battery-ac-adapter is a status icon

PR #609, opened 2015-07-03
Moves app status icons to new folders

wa4557:
Ok I've made a first approach to move the app status icons to new folders (`panel` in `16` `22` and `24`) I in order to solve #190. I tried to seperate the status icons from the app icons. I have not created or modified existing icons just moved them. Open for discussions!


dirtydancing:
Great! Just had a quick look, will come back to this later.

Question: Did you exclude the "toxic" status icons, i.e. those that have the same name as the proper (Circle) app icons? Cf. https://github.com/numixproject/numix-icon-theme/issues/192.

On first glance, I noticed that e.g. `vlc.svg` is among those moved to `panel`. These cannot be moved to `panel`, though, before being renamed. For the time being, these can remain in `status`, where they will do no harm to Numix-Circle (with `panel` to `panel` folder symlinks in Numix-Circle).

wa4557:
i'd suggest to split it up as little as possible so I think we should move them and afterwards rename them. We can of course move it and rename them and afterwards symlink it to circle

Foggalong:
From what I can tell there's a lot of icons in this pull that shouldn't be included in this merge. The cloud syncing clients, the workspace indicators, etc, are all status icons that arne't affected by this issue. Ideally we should *only* be moving apps which are part of this problem.

wa4557:
Well ok I thought that it makes sense to move all the app icons (why not?) to the new folders. If we just move those icons that do have the problem this might/will result in huge inconsistency. But ok if we just move the icons that cause the problems we might have to do a list: From my personal experience the problematic icons are:

redshift-on/off (those are status icons as well btw.)
variety
indicator-sound-switcher
disper-indicator
evernote/everpad
caffeine-* (status icons)

Foggalong:
The line between what's an "app" icon and what's a "status" icon is very blurred in this context. I'd for example class all of the filesync icons, the on/off type icons, the workspace indicators, language indicators, all as status icons because they display the status of the app. When it comes down to it, all the battery status icon is a indicator which shows the status of the battery manager (which is reading the battery).

Since this commit seemed to contain the majority of the contents of the status directory it might be easier to work conversely. If we were to move all the app icons, what would you class as _not_ an app icon and leave in status?

wa4557:
I'd say that an icon that has a corresponding icon in circle is an app icon. All the other icons are status icons :)

Foggalong:
I'd argue that there being a `dropbox-sync-status-01.svg` or whatever status icon though isn't the same as the `dropbox.svg` app icon because otherwise almost every status icon comes from an one with an app. The exception being the oddball apps (looking at you pushbullet and privacy indicators) which ship status icons and but no app icon but they don't really matter too much in this.

It's only some on/off icons through right, not all of them?

wa4557:
This definition is fine as well, imo. From my experience it's only caffeine
and redshift that are the only on/off icons

2015-07-03 13:53 GMT+02:00 Joshua Fogg <notifications@github.com>:

dirtydancing:
1) I would argue as well that all the "app status icons" that also have an app icon in Circle should be moved to `panel`.

imo this is a matter of practicality: we have seen that not all "app status icons" cause problems, which might have to do with the toolkit they use. But in order to figure out exactly which "app status icons" cause problems, all of these would need to be tested. And even then, all of these are moving targets (and can e.g. change their toolkit), so there is a danger of this becoming an unnecessary maintenance nightmare. And potentially, it is exactly the icons with Circle app icons that can cause problems.

2) I would furthermore argue that the "toxic" app status icons (we already have a list) should be kept in `status` and be renamed before being moved to `panel`.

Again, the reason is a pragmatic one: this way, all of this can be done step by step and does not have to be "finished" before the folder symlinks are added to Numix-Circle. Actually, I would recommend to add the `panel` to `panel` folder symlinks to Numix-Circle right away, to get this going and to allow for Numix-Circle PPA updates again. Of course, if all the renaming is very simple, this could all be done in one step.

3) As to reverting: imo it does not matter if there are some more or some less icons moved from `status` to `panel`: reverting would simply mean to take all the icons from `panel` and move them back to `status`.

Foggalong:
1. Moving them all now is a bit too much, especially when we don't know for certain if this is going to cause problems on certain desktop environments. For now it's just a case of moving stuff incrementally.

  As an aside, if they're changing toolkit that often involves switching to a completely different app indicator system which makes the point moot in that particular instance.

2. Keeping the same icon in two places isn't an option either. It makes things needlessly complicated maintenance wise and doesn't tell us what part of the solution is actually fixing things.

  All of the symlinks are already in Numix Circle and Numix Circle Light- they were fairly trivial to implement. The Circle PPA updates won't be set going until this issue is completely fixed which will hopefully be within a few days.. If this does turn out to cause problems then we want to know as few users as possible are being as affected.


3. It means bigger commits that are harder to verify and it clogs up the git history. Big commits in general are just a bad idea.

-----

Please continue this discussion in #190 - it helps keep the discussion on this topic all in one place and in somewhere more easily discoverable (since this pull is now closed).

PR #614, opened 2015-07-06
Moves several status icons to panel

wa4557:
Those are the icons I personally have problems with. 
For variety I had to create symlink from light to base because these icons did not exist in Numix-Light and Numix-Circle-Light does not inherit Numix-Circle


wa4557:
I had to revert the everpad/evernote move because those icons are toxic icons with the same name in base and circle :(

Foggalong:
It's finally all coming together :)

PR #533, opened 2015-06-06
Renamed several status icons

dirtydancing:
Addresses https://github.com/numixproject/numix-icon-theme/issues/192. Cf. conversation over there.

PR #623, opened 2015-07-08
Renames several bad status icons

wa4557:
I started working on renaming a few files that use the same icon name in `status` and `apps`. 

I will continue to add to this PR (one app per commit) until I'm finished, you tell me I should stop because it's stupid what I'm doing, or you think the PR is large enough (I will start a new one then).

I won't include the ones from #533 because this PR is just as good. If there were symlinks to those bad icons I changed the files the symlink was pointing to into regular files and changed the renamed file in a symlink instead...


wa4557:
That's it except. For the icons that are already dealt with in #533 and the bluetooth-icons. This are all the icons that need to be renamed

wa4557:
I'm extraordinary stupid? My previous PR killed this one :( I think I have to do that again

Foggalong:
Ah :( sorry!

PR #625, opened 2015-07-08
Corrects colour for several status icons

wa4557:
Several icons from `status`/`panel` still had the "wrong" colour `#dcdcdc`. I ran a simple grep and changed the colour to the correct and new `#ececec`


wa4557:
I'll close this, because a lot of these wrong colours are defined in unused definitions, so I have to run `inkscape --vacuum-defs` first. Still there are some icons that have the wrong colour

PR #626, opened 2015-07-09
Vacuum cleaning and correct colouring

wa4557:
When running `grep -r #dcdcdc /status` it turns out that a lot of icons still have the wrong colouring. In order to solve this issue I ran `inkscape --vacuum-defs` to remove all of the unused definitions (that's why  the PR is so big; but no other changes), and afterwards `sed -i "s/#dcdcdc/#ececec/g" /status/*.svg.

It turns out that some of the battery icons had the wrong colour, autokey (shame on me) and a few others


Foggalong:
dat diff

Issue #613, opened 2015-07-04
Qt apps indicator


bil-elmoussaoui:
Hardcode-Tray now support Qt applications using the patched version of sni-qt. 
I added a support to Skype,Megasync and Google Music Manager
For Skype : 
- [x] skype-status-connected.svg
- [x] skype-status-connected-notification.svg
- [x] skype-status-disturb.svg (for do not disturb status)
- [x] skype-status-disturb-notification.svg
- [x] skype-status-offline.svg
- [x] skype-status-offline-notification.svg
- [x] skype-status-invisible.svg
- [x] skype-status-invisible-notification.svg
- [x] skype-status-away.svg
- [x] skype-status-away-notification.svg
- [x] skype-status-connecting-0.svg
- [x] skype-status-connecting-1.svg
- [x] skype-status-connecting-2.svg
- [x] skype-status-connecting-3.svg

For Megasync 
- [x] megasync-status-paused.svg
- [x] megasync-status-disconnected.svg
- [x] megasync-status-syncing.svg
- [x] megasync-status-ok.svg

For Google Music Manager : 
- [x] google-music-manager-panel.svg

For Google Chrome:  
- [x] google-chrome-notification
- [x] google-chrome-notification-disabled
- [x] google-chrome-no-notification
- [x] google-chrome-no-notification-disabled

Those indicators won't affect #190 and so they can be added without any problem

wa4557:
You have a type in your list: distrub should read disturb (skype)

wa4557:
I guess they belong then to `status`?

bil-elmoussaoui:
@wa4557 Yeah

wa4557:
I'm kinda excited that it is now possible to theme the skype icon, I kind of avoided to use it, because it always looked so ugly in the tray :)

bil-elmoussaoui:
Numix everywhere :dancers:

bil-elmoussaoui:
I added google chrome indicator icons since it's possible to fix them using HardcodeTray

pgera:
Can you add some documentation about the hardcode script to the main README or to the README of the hardcode repo ? It is not very clear how the installation/removal is handled. Is it a one time call to the script ? How would one revert back to stock ? If this is a long term solution, would it make sense to create a separate package and add it to the ppa ?

bil-elmoussaoui:
[...]

PR #627, opened 2015-07-09
Renames several bad status icons

wa4557:
I was able to save most of my work (only a few conflicts \o/, git is really powerfull). I also added vlc, brasero and google-chrome, because @dirtydancing asked me to include them. 
I also ran `inkscape --vacuum-defs` again for those icons, because they were made before the last PR


dirtydancing:
Just putting in a reference to https://github.com/numixproject/numix-icon-theme/issues/192, so it is clear which issue this PR addresses, and tracked in https://github.com/numixproject/numix-icon-theme/issues/192 :)

Foggalong:
Just noticed there's the Chrome icons here. Would it make sense to instead move them to the icon names given in #613 so that they can be used?

wa4557:
Partly, (there are the notification icons as well, I guess they need to be coloured blue). I'd still keep these icons and make symlinks to those. If a user wants to revert this whole `renamed-` thing the icons are already present...

Foggalong:
I guess that makes sense, though we will eventually have to draw the line somewhere (when developers update their app to fix the problem for example).

wa4557:
Yeah true. I'll add the chrome icons when and if this PR is merged :)

Foggalong:
DO NOT DELETE YOUR REPO YET. If I need to revert this pull it's a lot easier if your repo is still in place

wa4557:
Roger. With this merges #190 and #192 are fixed :)

Foggalong:
Gonna do a little bit of testing with these fixes, and the a commit which adds the "install both in the same place" clarification to the README and in that commit I'll close those issues :)

pgera:
I don't see these icons in Unity's dash search any more.

dirtydancing:
@pgera Try an icon cache update (assuming you have Numix base installed globally):

`sudo gtk-update-icon-cache /usr/share/icons/Numix`

If you see no immediate change, to keep it simple: restart your machine. Any luck? :)

pgera:
@dirtydancing , No luck. Tried both.
screenshot from 2015-07-10 23 53 47

wa4557:
Close the dash; change to another icon theme (doesn't matter which one); change back to Numix; run `killall unity-scope-home`. 

I cannot reproduce your issue :(

wa4557:
what does `notify-send test -i google-chrome` give?

dirtydancing:
Just in case: try a forced icon cache update: `sudo gtk-update-icon-cache -f /usr/share/icons/Numix`

pgera:
Force update did it. Thanks!

dirtydancing:
:+1:

PR #638, opened 2015-07-16
Added document-import action icon

palob:
Added ``document-import`` (one from #437) by basically mirroring existing ``document-export``.
Fixed alignment in ``document-export``.

Exapmle (altered colour for the sake of visibility):
``document-export``
document-export

``document-import``
document-import


palob:
``document-import`` is on the #580 list as well.

PR #641, opened 2015-07-17
Added missing archive-... action icons

palob:
From #580

``archive-extract`` as symlink to ``tap-extract-to``
``archive-insert`` as symlink to ``tap-create``
``archive-insert-directory`` as symlink to ``add-folders-to-archive``

Pushed design for ``archive-remove``:
archive-remove
(true to original Oxygen icon)

Alternative:
archive-remove2
(in line with existing ``package-remove``)


Foggalong:
Go with the alt :)

palob:
Alt is up + a few alignment fixes to existing icons.

PR #642, opened 2015-07-17
Symlinks for the 'edit' action icon

palob:
One from #437.


palob:
The check is missing :)

PR #670, opened 2015-08-22
Replaces old styles by symlink

wa4557:
A few `xfpm-keyboard` icons still had the old design. I replaced them by symlinks and also added missing symlinks


Foggalong:
Thanks!

PR #675, opened 2015-08-29
adding icons for indicator-stickynotes

ezeeyahoo:


Foggalong:
We use SVG rather than PNG for our icons, sorry. If you have those to push I'd be happy to take a look at the

bil-elmoussaoui:
And you should add 16x16;22x22; 24x24 sizes instea of 48x48 ;)

wa4557:
If you want to add those:

1) Svgs for `16x16`, `22x22` and `24x24`, where the `22px` and `24px` icons have the same smbol size only a different page size. The `16px` icon is a scaled down version of the 22px one. All three sizes aligned to grid.

2) Colour of the icons in Numix `#ececec`

3) The same icons should be added to Numix-Light as well with colour `#353535` (which makes in total 6 icons)

4) where to put them: that's where it becomes a bit difficult. Depending if the app is a gtk icon they should be placed either in `status` or `panel`

PR #687, opened 2015-09-03
mail-open-multiple

wa4557:
Adds a png icon and deletes a fogotten png icon. I accidentely also merged te mail-unread multiple, I will therefore delete the other pull-request


Foggalong:
Ah, on see now :)

Issue #450, opened 2015-04-11
Non-Numix outadated png icons [$5]


bil-elmoussaoui:
Here's a list of some of icons in the numix icons that are png and not really looking like numix icons. 

 - [x] application-x-cue.png
 - [x] contact-editor.png
 - [x] folder-copy.png
 - [x] folder-move.png
 - [x] go-today.png
 - [x] im-aim.png
 - [x] im-icq.png
 - [x] im-jabber.png
 - [x] im-msn.png
 - [x] im-nov.png
 - [x] im-yahoo.png
 - [x] mail-copy.png
 - [x] mail-move.png
 - [x] offline.png
 - [x] online.png
 - [x] opera-extension.png
 - [x] preferences-autocompletion.png
 - [x] preferences-calendar-and-tasks.png
 - [x] preferences-certificates.png
 - [x] preferences-composer.png
 - [x] preferences-mail-accounts.png
 - [x] preferences-mail.png
 - [x] preferences-system-network-proxy.png
 - [x] query-free-busy.png
 - [x] stock_alarm.png
 - [x] stock_bell.png
 - [x] stock_check-filled.png
 - [x] stock_contact-list.png
 - [x] stock_insert-note.png
 - [x] stock_insert-rule.png
 - [x] stock_insert-table.png
 - [x] stock_mail-filters-apply.png
 - [x] stock_mail-flag-for-followup-done.png
 - [x] stock_mail-flag-for-followup.png
 - [x] stock_mail-open-multiple.png
 - [x] stock_mail-unread-multiple.png
 - [x] stock_new-24h-appointment.png
 - [x] stock_new-meeting.png
 - [x] stock_notes.png
 - [x] stock_people.png
 - [x] stock_score-high.png
 - [x] stock_score-higher.png
 - [x] stock_score-highest.png
 - [x] stock_score-low.png
 - [x] stock_score-lower.png
 - [x] stock_score-lowest.png
 - [x] stock_score-normal.png
 - [x] stock_select-column.png
 - [x] stock_select-row.png
 - [x] stock_shared-by-me.png
 - [x] stock_shared-to-me.png
 - [ ] stock_show-all.png
 - [x] stock_signature-bad.png
 - [x] stock_signature-ok.png
 - [x] stock_signature.png
 - [x] stock_task-assigned-to.png
 - [x] stock_task-assigned.png
 - [x] stock_task-recurring.png
 - [x] stock_task.png
 - [x] stock_text-monospaced.png
[...]

PR #689, opened 2015-09-04
Adds stock_mail-flag* icons

wa4557:
The last mail icons from #450. I just modifi an already very pretty circle icon


Foggalong:
Love it! Very creative solution

PR #690, opened 2015-09-04
Added stock_addressbook and stock_calendar action icons and evolution…

palob:
… action icons

So far ``evolution-contacts`` and ``evolution-calendar`` doesn't seem to be applied yet and mimetype icons are shown instead as discussed over in https://github.com/numixproject/numix-icon-theme-circle/pull/2518

After merging https://github.com/numixproject/numix-icon-theme-circle/pull/2526 as well it'll look better than before when app icons were shown anyway.

### DIFF
--- a/Numix/128x128/actions/evolution-addressbook.svg
+++ b/Numix/128x128/actions/evolution-addressbook.svg
@@ -1,0 +1,1 @@
+stock_addressbook.svg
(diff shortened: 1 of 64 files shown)


Foggalong:
Is it not possible to have `stock_calendar.svg` as a symlink rather than a copy?

palob:
It's a slightly different design without a day highlighted.

Foggalong:
Ah, my bad. I'd thought I'd seen one without the highlight but I guess not. Thanks for the pull!

PR #695, opened 2015-09-05
Adds preferences-desktop-accessibility

wa4557:
Starting with the low hanging fruits from #437 :)

### DIFF
--- a/Numix/128x128/categories/preferences-desktop-accessibility.svg
+++ b/Numix/128x128/categories/preferences-desktop-accessibility.svg
@@ -1,0 +1,61 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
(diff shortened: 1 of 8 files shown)


palob:
Isn't this an app icon? Used for a settings plug?

At least it's in Circle.

wa4557:
A Lot of cazegories icons are in circle. All of the prefernces icons are in categories

Foggalong:
Where does elementary actually show categories?

wa4557:
Idk but since the Icon is called prefernces- I guess somewhere where prefernces are present

Foggalong:
If it's in the preferences we aim to have them look like app icons. We tried letting them have a different style (inc. the same as categories) but it was ultimately a failure because too many apps use icons also shown in the settings as their app icons.


screenshot_2015-09-05_18-05-30

wa4557:
Yes, but there are already app icons, which means that the circle is used no matter what. That's just if base is used without circle

wa4557:
But I agree: If there are `preferences-` icons in those lists I will look beforehand If an icon in circle already exists, and if not I will add one in circle and base

Foggalong:
Ah, I see! Good thinking :)

PR #696, opened 2015-09-05
Added appointment-soon and appointment-missed status icons

palob:
From the #577 list.

Furthermore I fixed librsvg problems of the original ``appointment...`` design.

In case you didn't know @wa4557 @dirtydancing to avoid this issue two (or more) objects on top of each other may not have the same outline. Otherwise the lower object's outline will shine through due to a librsvg bug (or just the way it works).

This is how it looked when the inner white circle shared its outline with the outer rim:
notfixed

And this is how it looks after I shrank the inner ring and moved it up+right:
nowfixed

The effect is more prominent with smaller icon sizes.

A similar outline shows up when both objects (grey ring, white circle) don't overlap at all. 
The background shines through:
nooverlap


dirtydancing:
Thanks for pointing out this librsvg issue @palob!

PR #698, opened 2015-09-06
Added task-due and task-past-due status icons

palob:
From the #577 list.


Foggalong:
I really like task past due idea, kudos

PR #701, opened 2015-09-07
Added stock_check-filled

palob:
Evolution status icon. From #450.

Shrinkened and recoloured variant of ``dialog-ok``.


wa4557:
Cool, I actually really dislike the green colour that is used for all the other action icons, the one you used is better to the eye...

palob:
It's the one from for instance ``dialog-info`` in ``status``.
I don't like the Numix ``actions`` "green" either, it isn't even green really.

PR #709, opened 2015-09-11
Adds applications-education

wa4557:
from #437


palob:
On the #580 list as well.

Issue #61, opened 2014-01-20
Missing mail-archive icon


harp37:
A numix icon for "mail-archieve" (action) is missing.
The original icon is attached.
![mail-archieve](https://f.cloud.github.com/assets/1820674/1951535/7aa9dd8c-8183-11e3-9c33-2c868cb30c85.png)

PR #716, opened 2015-09-12
Added mail-archive and mail-mailbox

palob:
From #437.


bil-elmoussaoui:
This will fix #61 too

Issue #273, opened 2015-02-05
Clementine actions icons


varlesh:
![2015-02-05 14 01 38](https://cloud.githubusercontent.com/assets/8083855/6059337/656ad36e-ad41-11e4-973d-feb60d865717.png)
left to right:

* [x] folder-sound
* [ ] view-media-playlist
* [x] multimedia-player-ipod-mini-blue
* [ ] view-media-lyrics
* [ ] x-clementine-artist
* [ ] edit-clear-locationbar-ltr
* [ ] edit-clear-list

bil-elmoussaoui:
I can symlink folder-sound to folder-music , @Foggalong should i add a pull request here or in Numix Folders?

Foggalong:
@bil-elmoussaoui Folder symlinks go in base, only source icon files in Numix Folders :)

palob:
I suppose the last two of the list above are reversed as to "from left to right".
``edit-clear-list.svg`` as symlink to ``process-stop.svg``?
``edit-clear-locationbar-ltr.svg`` as symlink to ``edit-clear.svg``?

Foggalong:
Fixed in 6dd800c01db639f8b737fdb3588a735fdab9c2fe

PR #751, opened 2015-09-26
Add icons for clementine

varlesh:


Foggalong:
Thanks for the contribution, is this for fixing #273? There are a few things that need changing though:

* All the icons & symlinks need to go in all the sizes
* The symlink you chose for `edit-clear-locationbar-ltr.svg` doesn't fit with the existing icon for `edit-clear-locationbar-rtl.svg`
* `x-clementine-artist.svg` should be a symlink rather than a SVG
* `view-media-playlist.svg` and `view-media-lyrics.svg` feel like they'd be better suited to being standalone icons with a bit more detail.

varlesh:
Why all size? Used only 16x16 and 22x22
I don't know why developers used edit-clear-locationbar-ltr. I know edit-clear-locationbar-ltr have other style (see symbolic icon), but this icon not good looking on clementine.

x-clementine-artist - i don't find analog icon and create new icon. Exist only categories/preferences-desktop-personal-directory, but this icon have another size 10x11

symlink - it's easy way )) Maybe create new icons for playsit and lyrics.

Foggalong:
* Larger sizes are used in HiDPi, XHiDPi, etc and could also be used by other applications
* If the name & context indicates the icon should be a certain way then that's the way we do it. If you think the Clementine develops picked the wrong icon though then report it to them and they might change it
* Ah, in that case okay.
* Symlinks are easier yeah, but why do it if it doesn't make sense in the context?

varlesh:
Again HDPI, i can not hide from it! 
If anybody create new icons for playlist and lyrics I'm not against.

varlesh:
for lyrics?

varlesh:
or this

Foggalong:
The latter would be good :)

varlesh:
for playlist?

varlesh:
or this

varlesh:
Also change multimedia-player-ipod-mini-blue (from yourself).
final result

varlesh:
I commited, you decide.

varlesh:
If all ok, HDPi support add later

Foggalong:
Of the two playlist icons, I kinda prefer the first. Yeah, happy to wait for HiDPi before merging :)

varlesh:
@Foggalong Please check

Foggalong:
Isn't the `ltr` icon the wrong way around?

varlesh:
You said earlier that you do not like symlink edit-clear-locationbar-ltr to dialog-close. I create new icon, variation edit-clear-symbolic.

varlesh:
This ltr
ltr
This rtl
rtl

Foggalong:
Yeah, the icon names are pretty much the wrong way around. As you can see in your original screenshot clementine uses the bottom one while you pushed the top. Unless I've misunderstood? But yeah that's the general design to go for :)

varlesh:
I suggest symlink edit-clear-locationbar-ltr to dialog-close, this better variant.
You agree?

Foggalong:
No, for reasons mentioned above :/

palob:
I think in the ``clear-locationbar`` case it's right that the ``rtl`` icon is (apparantly) the wrong way around.
The icon is used in a wrong way by Clementine. It is supposed to be at the and of a text entry field pointing against the direction you write.

As far as I understand here ``rtl`` or ``ltr`` means the direction your language is written (e.g. ``ltr`` for English, ``rtl`` for Arab).

palob:
@varlesh I've had a look at your branch. All icons are perfectly aligned to the grid so the PR is ready to be merged from this perspective.

I'm still a bit unsure about the ``rtl/ltr`` thing.
In accordance to my explanation above the arrow in ``edit-clear-location-bar-rtl`` is supposed to point to the right and vice versa (``rtl/ltr`` doesn't refer to the arrow direction).

This is how it looks in other themes:
+ Adwaita
edit-clear_adwaita

+ elementary
edit-clear_elementary

+ Breeze (``rtl`` **seems to be wrong here**)
edit-clear_breeze

As I understand it the icon is supposed to be at the end of the text entry field.
Clementine however uses it at the beginning.

@varlesh Is the icon Clementine uses really ``edit-clear-location-bar-rtl`` instead of ``edit-clear-location-bar-rtl``?

Anyway, in order not to look awry the icon should point to the right. Maybe it's okay to replicate the mistake, Clementine might be the only program using it.

varlesh:
Realy use edit-clear-location-bar-ltr, please see source:
https://github.com/clementine-player/Clementine/blob/master/data/icons/22x22/edit-clear-locationbar-ltr.png
I don't know why developers clementine use ltr icon and located on the left.
I not seen this icon anywhere else...

PR #759, opened 2015-09-27
mail-encrypted action icon

palob:
From #580.

9th changed file is alignment correction.


wa4557:
Hmm. Shouldn't the mail envelope centered vertically?

wa4557:
Nevermind forget it, all the other icons are the same

Issue #664, opened 2015-08-12
Inconsistency in device icons depending on Ubuntu icon size settings


keunes:
Testing things for another bug with the launcher icons, I discovered that when I set the icon size 40 or higher, on my external monitor the icon for my windows partition is not the icon that I'm used to.

I reported a [bug](https://github.com/numixproject/numix-icon-theme-circle/issues/2476) for Numix Circle, but apparently it's an issue with the base theme. Would expect icons to remain consistent, regardless of size settings :)

Expected (38 & under) and unexpected (40 and up) icon:
![screenshot from 2015-08-12 22 14 37](https://cloud.githubusercontent.com/assets/11229646/9235595/eb420b14-413f-11e5-8022-8220e57455cf.png) ![screenshot from 2015-08-12 22 13 29](https://cloud.githubusercontent.com/assets/11229646/9235594/eb25c2ec-413f-11e5-9ee7-c6ff4c2924fa.png)

Foggalong:
I'll fix this asap :)

wa4557:
What's the way to fix this? Adding polychrome icons to larger sizes as well? (Probably just a scaled up version of the `32x32` icons)

Foggalong:
Yeah, I need to ask someone why the devices icons above 48px are monochrome and check it is intact just a case of never being updated.

wa4557:
Fortunately if we are going to add polychrome icons for larger sizes this might be a very easy task. Almost all the icons for lower sizes are nicely aligned to grid

Foggalong:
Excellent :D

wa4557:
Obviously larger icons just weren't added. Looking at [this] (https://github.com/numixproject/numix-icon-theme/commit/7f88cbeefe4dc47c0fb8f8e1ded0c6003a3ae041) commit only icons up to `48x48` were added

wa4557:
https://github.com/holylobster/nuntius-linux#the-icon

sljunkie:
Devices icons in /usr/share/icons/gnome (Unity, Cinnamon and Gnome-shell use them):

$ pwd
/usr/share/icons/gnome/256x256/devices
$ ls -l
total 1416
lrwxrwxrwx 1 root root    16 Apr 15 18:38 3floppy_unmount.png -> media-floppy.png
-rw-r--r-- 1 root root 35206 Apr 15 18:38 ac-adapter.png
[...]

PR #763, opened 2015-09-27
Fixes wrong size for image-loading in 96px

wa4557:
Title says it. Btw: @Foggalong You've added your latest action icons also to 96px, is this something you want?, I'm just asking because as far as I can see all the icons (except the places icons of course) in 96px are just the same icons as in 48px just multiplied by two. This makes this a bit redundant and could for example also be defined in the index.theme (same is also true for 64, 128 and 256 btw)

### DIFF
--- a/Numix/96x96/places/gnome-fs-loading-icon.svg
+++ b/Numix/96x96/places/gnome-fs-loading-icon.svg
@@ -7,12 +7,12 @@
    xmlns="http://www.w3.org/2000/svg"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="48"
-   viewBox="0 0 48 48"
-   height="48"
+   width="96"
+   viewBox="0 0 96 96"
+   height="96"
    id="svg2"
    version="1.1"
-   inkscape:version="0.48.5 r10040"
+   inkscape:version="0.91 r"
    sodipodi:docname="gnome-fs-loading-icon.svg">
   <metadata
      id="metadata12">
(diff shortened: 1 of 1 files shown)


palob:
I've wondered before too whether we really need all the sizes in case they're just scaled. 
By the same token can say 64px icons even show up in place of a 24px icon o HiDPI screens?

That said a few of my icons use a distinct design for 128px and 256px (not crucial to preserve however).

wa4557:
The question is how they are scaled up. Let's say we have icons up to 64px (which would be reasonable imo) and the app wants to use a 96px icon. The best way would be to use the 48px icon and multiply it by two. If however the 64px icon is used and multplied by 2/3 this would be less ideal,. 
If ths happens it would be better to have all the sizes

Foggalong:
As I understand, the reason for having a 96px size is that if an app is requesting it and it's not there it'll go to the next nearest size which in our case would be 64px or 128px which would cause blurring. GNOME recently brought 96px to the forefront where as previously it hadn't been touched on so much.

If there's a way to reliably replace the icons with x2 versions though just using the index.theme variable though that would be awesome since we could knock out everything about 64px (I think the largest we have with distinguishable features?)

wa4557:
There's definitely a way to do this, I'll have to play around a bit. for everything except `places` we could remove sizes bigger than `64px`. That would be extremely cool, because maintenance wouldbecome way easier. 

Anyways I'll open a new issue about that, as soon as I found out something, which means that in the meantime this PR can be merged.

sljunkie:
GNOME uses 'big' 256x256 icons... see https://github.com/numixproject/numix-icon-theme/issues/664

wa4557:
@sljunkie Working on it, see #767

PR #769, opened 2015-10-04
Adds support for 96x96

wa4557:
A big push :) I also changed `gnome-spinner` in `animations` to a symlink because it was an identical file


wa4557:
I'll keep this one open, but I actually suggest to merge the other PR :)

wa4557:
I'll close this one. Adding more files to the theme is arguably not the way to fix this. My other PR is a waay better way to fix the 96px size issue

PR #770, opened 2015-10-04
Alignment fixes for emblem icons

palob:
and symlink replacemnt for redundant designs

There are still lots of misalignments left, I've only fixed the most blatant ones.

Sizes above 64px are definitely just scaled up.


wa4557:
Great! I think the emblems were the last ones... Everything else should be up to date.

PR #771, opened 2015-10-04
Adds hidpi support and removes unnecessary files larger than 64px

wa4557:
Ok this is the other approach where all sizes > 64px are dropped. I kept the icons from `places` and `actions` for larger sizes, because they are distinct in multiple ways, but I did not keep the other folders because they were pretty much the same (there were just a few minor differences, which I think we can neglect for the sake of consistency, (@palob your call, if there's something you really want to keep say it and I can add it).

The easiest way to test that it really works is to open an `ipython` console changing to this icon theme and running:

```
from gi.repository import Gtk
icon_theme = Gtk.IconTheme.get_default()
icon_theme.lookup_icon("gtk-edit", 96, 0).get_filename()
```
and as you will see the icons from the newly created folders are used :)
If you change size/name you can play around, so far all the icons were picked from the correct location


palob:
What exactly made you keep the ``actions`` folders?

wa4557:
Because there are some icons from `places` (like `fileopen` etc. ) . But I just kept those icons, the other ones are from `48px` and `64px`

palob:
OK, this makes the theme a lot more lightweight. We'll only lose slight improvements for bigger sizes (like bigger symbols on mimetype icons, realignments, relatively smaller outlines..) as far I can see.

The ``elementary`` theme has a 128px size for some of it's sizes.
Maybe we should ask an *icon theming luminary* to be sure we're safe?

wa4557:
You think it's ok if we drop them? I could also add specific folders/files if you think they are necessary...

Foggalong:
@palob I don't think there are any symbol size differences between 64px and 128px: 64px is the last size to have distinctions

palob:
Well, I did design a few 128/256 icons distinctly but I think it's best to toss them in order to keep it simple.

Foggalong:
Just as a note on the `16x16` to `16`, that change will have to be applied to Numix Circle and Numix Folders simultaneously so that they all work as expected :)

wa4557:
I already pushed to circle, will push to folders asap

wa4557:
Proof that it works: 

Program used: 
```
from gi.repository import Gtk, Gdk, GdkPixbuf, Gio

class HelloWorldApp(Gtk.Application):
    def __init__(self):
        Gtk.Application.__init__(self, application_id="apps.test.helloworld",
                                 flags=Gio.ApplicationFlags.FLAGS_NONE)
        self.connect("activate", self.on_activate)
        
    def on_activate(self, data=None):
        window = Gtk.Window(type=Gtk.WindowType.TOPLEVEL)
        window.set_title("Gtk3 Python Example")
        window.set_border_width(24)
        sizes = [16,22,24,32,48,64,96,128,256]
        vbox = Gtk.HBox()
        for size in sizes:
            fixed = Gtk.Fixed()
            image = Gtk.Image()
            icon_theme = Gtk.IconTheme.get_default()
            name = 'gtk-edit'
            pixbuf = icon_theme.load_icon(name, size, Gtk.IconLookupFlags.USE_BUILTIN)
            print(icon_theme.lookup_icon(name, size, 0).get_filename())
            image.set_from_pixbuf(pixbuf)
            vbox.pack_start(image, True, True, 10)
        window.add(vbox)
        window.show_all()
        self.add_window(window)
    
if __name__ == "__main__":
    app = HelloWorldApp()
    app.run(None)

```

With new system:
output: 
```
/home/andreas/.icons/Numix-Test/16/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/22/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/24/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/32/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/48/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/48@2x/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64@2x/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64@4x/actions/gtk-edit.svg
```
Screenshot:
bildschirmfoto vom 2015-10-05 16 52 21


old system: 
output: 
```
/usr/share/icons/Numix/16x16/actions/gtk-edit.svg
/usr/share/icons/Numix/22x22/actions/gtk-edit.svg
/usr/share/icons/Numix/24x24/actions/gtk-edit.svg
/usr/share/icons/Numix/32x32/actions/gtk-edit.svg
/usr/share/icons/Numix/48x48/actions/gtk-edit.svg
/usr/share/icons/Numix/64x64/actions/gtk-edit.svg
/usr/share/icons/Numix/256x256/actions/gtk-edit.svg
/usr/share/icons/Numix/128x128/actions/gtk-edit.svg
/usr/share/icons/Numix/256x256/actions/gtk-edit.svg
```
(96px was a scaled down version of 256

screenshot:
bildschirmfoto vom 2015-10-05 16 52 43

wa4557:
One more thing: If we need icons that are bigger than 256px, we have to add also 64@6x files, I guess for now this is not necessary, just something we have to keep in mind for future issue reports...

palob:
It looks like some status icons from ``24x24`` are now in ``16``. Is this on purpose?

palob:
By the way, does anybody know what the ``Type=scalable`` thing from [elementary](http://bazaar.launchpad.net/~elementary-design/elementaryicons/trunk/view/head:/index.theme) does and why ``MinSize=8`` for all sizes?

wa4557:
Which icons?

palob:
Mhh, maybe it's just GitHub messing up the diff again just as with ``+0 -0``. 
When I clone your branch I don't see any 24px in the ``16`` folder.

wa4557:
Ok although this PR solves the greater than 64px issue it is not a proper implementation of HIDPI support. Nevertheless this PR can and should be merged since it solves the afrorementioned 96px size issue. The good news is (and this is also related to @palob s question): I know how to proper implement hidpi support, and if this PR is merged I'm more than happy to provide that :)

palob:
I guess the always ``Min=8`` thing is for cases like for instance when you have a ``24px`` icon but no ``16px`` one. We might need that if we decide to design 16px, 24px, 32px Circle icons.

Regarding this PR, I think it should be merged after all other open ones which otherwise will face maerge confliczs.

wa4557:
Yes i think that's what the minsize entry is for. Yes I know that I will have to rebase (which is always a big pain) so maybe wait with new pull requests until this one is merged?

wa4557:
I fixed all the merge conflicts that came up because of the merged Pull requests, but I'd prefer to not do that again, because it's a pain in the ass. So it would be really cool if this one could be reviewed and merged...

Foggalong:
Sure thing, I just wanted to get the already existing pull requests merged so it was just your 2 left. I'll review this one asap :)

Foggalong:
Okay, starting to review this now. Some initial thoughts:

1. Is there a reason you've symlinked all the individual context folders rather than the size folder as a whole?

2. Am I right in thinking the replacements are as follows `96 -> 48@2x`, `128 -> 64@2x`, and `256 -> 64@4x`?

3. It's possible that there's going to be a naming conflict by the above mentioned replacement folders and the ones needed for full HiDPI support. Full support requires every folder of every size to have a `@2x` equivalent which means that for example the current `96x96` folder would need to have one called `96@2x` under this system, but by the naming convention used here it would end up being called `48@2x@2x` which is a bit of a mouthful.

wa4557:
1. That's because I don't wanted that all folders (including places) are scaled up, to the new size. I agree however that this is actually a bad idea when implementing proper HIDPI support, because there ALL the folders need a @2x equivalent. For proper HIDPI support this needs to be changed, and actually in my recent HIDPI branch I did that as you suggested. I'll open another PR for this as soon as this one is merged. For the 96px issue this approach is enough though

2. Yes correct.

3. Again, this is not a HIDPI implementation yet (I mentioned that in a previous comment), it just solves the `96px` size issue (see screenshots above), and for now this just means that there is no HIDPI for `96px` places. I have however full HIDPI support ready, so as soon this PR is merged I'll open another PR. I just didn't want to add that to this one because, it doesn't really fit. I agree that for full HIDPI support there has to be a `96@2x/places` folder.

Foggalong:
Goodo :+1: I don't run the relevant system to test all the intricacies of this pull so maybe hold on deleting the branch just now so we can easily revert if things go awry.

wa4557:
i'm prettty confident that all is fine :) one thing though: the numix circle ppa package needs to be upated otherwise the symlinks break

PR #777, opened 2015-10-14
Adds proper hidpi support

wa4557:
This is now a proper implementation of HIDPI. I removed the 64@4x folders because the same thing can also be done by using the MinSize and MaxSize variables correctly, and there is now a 2x folder for each folder (For 5k displays there needs to be 3x folders at some point). 

96px 128px and 256px icons (except places and some action icons) are still taken from 48@2x and 64@2x, so nothing changed there


Foggalong:
@wa4557 A few points:

1. I believe this is missing the 96@2x size

2. I think there's some error in here - with the current setup we'll be missing all the icons that aren't action/places for size 256 and possible 128 & 96.

3. Why the change from `Type=Fixed` to `Type=Scalable`?

wa4557:
1. True, I'll add that folder, but in the index.theme it is present
2. I disagree.The icons are still there and the icons should be picked from the correct locations. That's alos related to your 3. issue, that's what the MinSize, MaxSize variables are doing, since the maxsize of the 64px icon is used for 128px/256px and 48px is used for 96px (that's what we wanted because there are no differences in icons above 64px). 
3. Because it is a bit more flexible. If there are some missing icons (although I agree that Numix is quite complete) holes are filled.

Foggalong:
1. :+1:
2. Won't using the 64px ones with varied MaxSize variables mean the 96px icon are referred to first since they'll be closer to the desired size.
3. Cool! Just checking with @satya164 that there's not [a reason](https://github.com/numixproject/numix-icon-theme/commit/a16d6edef20604dc1f66ee7e26f0636099b8332a) we don't use it

wa4557:
1. Added
2. It doesn't. I have to think about why exactly but the correct sizes are used: 

```
/home/andreas/.icons/Numix-Test/16/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/22/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/24/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/32/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/48/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/48@2x/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64@2x/actions/gtk-edit.svg
/home/andreas/.icons/Numix-Test/64@2x/actions/gtk-edit.svg
```
sizes [16,22,24,32,48,64,96,128,256]

bildschirmfoto vom 2015-10-14 17 54 33

Foggalong:
It isn't that the icon aren't coming out the right size it's that the lookup might not be looking in the right place for them. e.g. looking for 96px but none there so goes to the nearest which is 64px rather than the scaled up 48px.

wa4557:
But they are picked from the correct locations, see the locations above the screenshots...

Foggalong:
Wait, were the icons for the render chosen via GTK lookup?

wa4557:
yes

Foggalong:
THAT'S SUCH A GOOD IDEA. OMG THAT'S BRILLIANT.

Foggalong:
I thought you'd just made a nicely rendered PNG of them all side by side, I didn't realise that was part of an interface you'd made for testing :D That's such a good idea! So much love kudos and thanks for you right now

wa4557:
Well thanks, not that hard actually :) For future reference: 

```
from gi.repository import Gtk, Gdk, GdkPixbuf, Gio

class HelloWorldApp(Gtk.Application):
    def __init__(self):
        Gtk.Application.__init__(self, application_id="apps.test.helloworld",
                                 flags=Gio.ApplicationFlags.FLAGS_NONE)
        self.connect("activate", self.on_activate)
        
    def on_activate(self, data=None):
        window = Gtk.Window(type=Gtk.WindowType.TOPLEVEL)
        window.set_title("Gtk3 Python Example")
        window.set_border_width(24)
        sizes = [16,22,24,32,48,64,96,128,256]
        vbox = Gtk.HBox()
        for size in sizes:
            fixed = Gtk.Fixed()
            image = Gtk.Image()
            icon_theme = Gtk.IconTheme.get_default()
            name = 'gtk-edit'
            pixbuf = icon_theme.load_icon(name, size, Gtk.IconLookupFlags.USE_BUILTIN)
            print(icon_theme.lookup_icon(name, size, 0).get_filename())
            image.set_from_pixbuf(pixbuf)
            vbox.pack_start(image, True, True, 10)
        window.add(vbox)
        window.show_all()
        self.add_window(window)
    
if __name__ == "__main__":
    app = HelloWorldApp()
    app.run(None)
```

Foggalong:
Thanks :heart:

wa4557:
Now that I think about it, this script would have been useful way earlier (for the infamous #190 for example), so @dirtydancing if you want to debug things like numix-icon-theme-circle#2601 this might be useful as well

Foggalong:
For future reference I've [added it here](https://github.com/numixproject/numix-labs/blob/master/Andreas/tester.py)

dirtydancing:
@wa4557 Wow, looks like an interesting thread. Maybe some day I will understand what exactly this is about :) Anyways, I am sure that you guys will know when to put this script into action for future issues.

PR #778, opened 2015-10-14
Adds new design for devices in 64

### LINKED ISSUE #664: Inconsistency in device icons depending on Ubuntu icon size settings

wa4557:
Adds a consistent design for devices icons in `64`. This is now necessary because the virtual `96px` devices icons are scaled up versions from 48. Fixes #664. Also aligned a few other icons to grid


bil-elmoussaoui:
Thank you so much! will make my desktop look better now! :dancer:

PR #782, opened 2015-10-15
Makes icons non-executable + simplification for clementine-indicator

wa4557:
Removed the executable but + I removed unnecessar clipping from clementine indicator icons


Foggalong:
:+1:

Issue #563, opened 2015-06-10
Mimetype: text/x-gettext-translation


hlechner:
Name: **translation file**
Mimetype: **text/x-gettext-translation**
Possible extensions: **.po**

![image](https://cloud.githubusercontent.com/assets/11244067/8094176/b45579f4-0f9b-11e5-9247-b0c70744586f.png)

hlechner:
Additional Info: I found on Oxygen theme the following Mimetype:
- text/x-po

It seems to be the same think, so a symlink to this one could be useful too.

palob:
Not easy to tackle, here are some raw drafts I'm not overly confident of:
![text-x-gettext-translation-48b](https://cloud.githubusercontent.com/assets/7050624/8267385/04750bc2-1760-11e5-836e-013222e0711e.png)
![text-x-gettext-translation-48c](https://cloud.githubusercontent.com/assets/7050624/8267386/04765cf2-1760-11e5-8cea-d9bd82da5ea9.png)
![text-x-gettext-translation-48d](https://cloud.githubusercontent.com/assets/7050624/8267387/047700bc-1760-11e5-94ea-3d060d2dfa00.png)

Any ideas?

Foggalong:
The first and second are okay because they link back to their app icon counterparts, but as you say not 100% on either

palob:
A new approach (nothing groundbreakingly new however):

``text-gettext-translation``/``text-x-po``
![16-64-po](https://cloud.githubusercontent.com/assets/7050624/10561129/2fbdfa06-751f-11e5-8e83-b92e8565ebdb.png)


``text-x-gettext-translation-template``
![16-64-pot](https://cloud.githubusercontent.com/assets/7050624/10561130/339e950e-751f-11e5-89cf-91822deb9fc3.png)

Foggalong:
I like the colourful implementation but it sort of feels like the flag isn't aligned properly. Maybe try 1px to the left?

wa4557:
I'd make the flag 1px narrower...

palob:
How about the non-dog-ear shape for the colourful ``.pot`` mimetype to be in line with the other *template* mimetypes?

@wa4557 So 1px narrower and 1px to the right?

Foggalong:
@palob I was forgetting about that, definitely do that

palob:
With or without the *PO* *POT* text?

wa4557:
I would keep the text

Foggalong:
[...]

PR #787, opened 2015-10-17
Added symlinks for .mo file mimetype

palob:
``application-x-gettext-translation`` (not to be confused with ``text-x-gettext-translation`` for ``.po`` files)

``.mo`` files are "compiled" gettext translation files. The files are binary so the visual representation with a symlink to ``application-octet-stream`` is better than the generic text mimetype icon.

### DIFF
--- a/Numix/16/mimetypes/application-x-gettext-translation.svg
+++ b/Numix/16/mimetypes/application-x-gettext-translation.svg
@@ -1,0 +1,1 @@
+application-octet-stream.svg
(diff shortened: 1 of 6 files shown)


Foggalong:
@palob Possibly use the colour [from here](https://github.com/numixproject/numix-icon-theme/issues/563#issuecomment-148949543) with the symbol from `application-octet-stream` to make the icons more unique?

palob:
Alright, will do all in one PR.

Foggalong:
:+1:

PR #794, opened 2015-10-18
Updates design of google-icon

wa4557:
Well since google updated their icon the old "google-g" doesn't really fit anymore


Foggalong:
:+1:

PR #796, opened 2015-10-18
Added application-x-abiword mimetype

palob:
From #580.


palob:
While doing this one I realised that many filetype extension mimetype icons suffer from non-ideal alignment.

I'll work on that.

PR #820, opened 2015-10-26
Adds a lot new scalable status icons

wa4557:
Got it sorted out, these are now a lot of new scalable status icons (and a few redundant changes on other icons with no effect, sorry about them, couldn't get rid of them reliably)


Foggalong:
Don't worry about the redundancies :)

PR #830, opened 2015-10-31
Small fix to index.theme

wa4557:
Because the places folders are special

### DIFF
--- a/Numix/index.theme
+++ b/Numix/index.theme
@@ -731,7 +731,7 @@ Size=48
 Scale=2
 Context=Places
 MinSize=8
-MaxSize=127
+MaxSize=63
 Type=Scalable
 
 [48@2x/status]


Foggalong:
Wouldn't actions be the same too?

wa4557:
Nope, actions are just like all the other folders. Except that there are also bigger sizes, but they do not interfere...

Foggalong:
Ah, I think I'm misunderstanding what this one does then

wa4557:
This is because for `places` `48` doesn't need to cover `96` because there is a seperate and complete folder for this size. Thus `48` only needs to cover sizes up to `63px`

Foggalong:
We have the same for actions with the 96px folder too, no? Does it not matter that even of few such icons exist?

wa4557:
But for actions there are just a few icons in the `96px` folder (the icons with folders in it), all the other icons should be scaled up versions from `48`. This means we cannot change the maxsize variable for these folders

wa4557:
Actually the fix in this PR is just a cosmetic one, because no matter what this variable is, the correct icon is used

Foggalong:
Ah, okay :)

PR #836, opened 2015-11-01
More new icons for Breeze

palob:
Some from #580.

+ edit-rename
+ shear
+ transform-move
+ view-list-tree
+ view-right-close
+ view-right-new
+ view-split-left-right
+ view-split-top-bottom

new designs:
kde


Foggalong:
I think there's some errors in the pull request. A few of the icons are showing up as non square and then also with irregular dimensions (like 16.0010px)

palob:
Looks like the 16.001px size originates from the original ``window-new.svg``. I'll fix it.

palob:
Should be fixed now. Only 16px sizze were affected, all other changes are tiny cleanups.

I couldn't find any non-square canvas.

Foggalong:
:+1:

PR #844, opened 2015-11-07
Next batch of breeze symlinks/icons

palob:
Some are on the #580 list.

+ gtk-undelete-ltr
+ list-add-user
+ list-remove-user
+ resource-group
+ resource-group-new
+ upindicator
+ user-group-new
+ user-identity
+ viewimage
+ view-media-artist
+ view-process-all
+ view-process-tree
+ view-process-own
+ view-process-system
+ view-process-tree
+ view-process-users

### DIFF
--- a/Numix/16/actions/gtk-undelete-ltr.svg
+++ b/Numix/16/actions/gtk-undelete-ltr.svg
@@ -1,0 +1,1 @@
+gtk-undelete.svg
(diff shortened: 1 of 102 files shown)


Foggalong:
As I understand `gtk-undelete-ltr` needs to be a mirror of the original icon, not just a symlink

palob:
``rtl`` is right actually, ``ltr`` (in breeze too) needs to be mirrored, doing this right now.

palob:
Ready.

PR #854, opened 2015-11-08
Mimetype icons for CoffeeScript files

palob:
Quite minimalistic design this time, in line with the JS mimetype.
allsizes


Foggalong:
The 16px icon is always gonna look a little bit off, but imo this looks slightly less off than the pushed copy
screenshot_2015-11-11_16-37-55
screenshot_2015-11-11_16-37-42

palob:
Now it looks like above :+1:

Foggalong:
:+1:

PR #857, opened 2015-11-10
Bigger 16px changes-allow/prevent

palob:
Looked quite tiny.


wa4557:
I like how inkscape becomes more and more numixy :)

PR #866, opened 2015-11-14
chronometer action icon

palob:
From #580.


Foggalong:
I had to look up what a chronometer was :P

palob:
Actually I think it's a chronograph in the icon rather than a chronometer :D

PR #873, opened 2015-11-16
Distinct APK mimetype icons

palob:
+ alignment fixes for the 64px DEB and RPM mimetypes.

Introducing a new letter: **K** :)


Foggalong:
I didn't realised we didn't already have one D: and oooooooh, that's why you wanted the font :package:

PR #883, opened 2015-11-19
More chronometer action icons

palob:
From the breeze theme.
chronometer-icons

### DIFF
--- a/Numix/16/actions/chronometer-lap.svg
+++ b/Numix/16/actions/chronometer-lap.svg
@@ -1,0 +1,95 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   xmlns:dc="http://purl.org/dc/elements/1.1/"
+   xmlns:cc="http://creativecommons.org/ns#"
+   xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+   xmlns:svg="http://www.w3.org/2000/svg"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
+   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
+   width="16"
+   height="16"
+   viewBox="0 0 16 16"
+   id="svg2"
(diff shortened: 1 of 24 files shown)


wa4557:
Since you are doing so much breeze icons lately I'm trying right now to add the missing breeze icons to the kde issue. Next week I might find some time to contribute icons again :)

wa4557:
Yeah I don't know if we should do that :( the number of missing icons is 2801...

palob:
Much appreciated :) By the way Oxygen will be [developed further on](https://kdeonlinux.wordpress.com/2015/11/08/breeze-oxygen-and-framework/) as ``oxygen-icons5``.

I've done some testing with Plasma 5 lately and Numix icon look pretty slick there.

palob:
Ooof! Hmm, I only see a total number of 1080 files in ``actions/22`` there.

wa4557:
So you suggest to only add the action icons?

palob:
As far as I can see the 2801 icons can't originate from non-action icons. 
breeze is like 85% action icons.

wa4557:
I'm taking the icons from https://github.com/NitruxSA/breeze-icon-theme/. Is this the right place?

palob:
Currently I'm examining the installed theme from the Arch repos.
I've posted a link in #580 to the 5.4.2 release version. 
Maybe there was a big contribution meanwhile?

wa4557:
Ok I was wrong about that. Compared wrongly, I'll add the icons right away :)

BTW: If Numix comes to KDE more complete Plasma 5 might be worth considering, looks pretty fine imo

palob:
Unfortunately there's a (hardly fixable) little issue with out-of-place Numix Circle icons in Plasma (not as out of place as in say Evolution however).

palob:
Browsing breeze's LibreOffice theme.... wtf how many icons does LO use?

wa4557:
That was exactly the problem, I compared the libreoffice icons as well, but I guess this is another kind of issue. 

Ok the correct number is now 596 icons, excluding the ones already reported in that issue. I hope a lot of them are symlinks ...

wa4557:
Ok done

palob:
:+1: Thanks!

PR #884, opened 2015-11-19
Symlinks for korg-todo and x-clementine-shuffle

palob:
2 from #580.

Action icons for Korganizer and Clementine.


Foggalong:
I wonder why `x-clementine-shuffle` is formatted like a mimtype when it clearly isn't?

Issue #768, opened 2015-10-04
Missing geary icons


bil-elmoussaoui:
Hello,
There's still some missing icons in Geary mail client.

- [ ] tag
- [ ] tags

Symbolic icons : 
- [ ] accessories-text-editor-symbolic
- [ ] application-menu-symbolic
- [ ]
- [ ]
- [ ]

wa4557:
Do you have a screenshot of that (how do for example tags and tag differ?)

bil-elmoussaoui:
@wa4557 I don't have those icons in Ubuntu Gnome! :S

palob:
So they're only used in the elementary UI?

bil-elmoussaoui:
Non! The version of geary shipped in Ubuntu Gnome uses symbolic icons

palob:
Ah OK, so that's the normal headerbar icons.

bil-elmoussaoui:
Yes, i will add the symbolic icons too

bil-elmoussaoui:
@wa4557 tags icon does not exists in the default elementary theme! maybe just having multiple tag icons? 
default tag icon in elementary
![Uploading Screenshot from 2015-10-31 00:01:13.png…]()

bil-elmoussaoui:
I updated this one! i added Geary Gnome icons (uses only symbolic ones)

palob:
Is ``tags`` the same as ``tag`` visually?

bil-elmoussaoui:
Yes

palob:
Hmm, so we can't just use the existing design from ``tag-new``, make it bigger and align it.

bil-elmoussaoui:
@palob You mean we can use the `tag-new` design ?

palob:
After a bit of tweaking. But no, because it's white.

There's already the issue with for instance ``document-import`` showing up as arrow only in some elementary apps' welcome screens when using the elmentary GTK theme.

PR #897, opened 2015-11-21
two symlinks for geary

bil-elmoussaoui:
#768  
- close-symbolic
- format-text-remove-symbolic

### DIFF
--- a/Numix-Light/scalable/actions/close-symbolic.svg
+++ b/Numix-Light/scalable/actions/close-symbolic.svg
@@ -1,0 +1,1 @@
+window-close-symbolic.svg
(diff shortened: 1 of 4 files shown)


palob:
Do you know where ``close-symbolic`` is used?

bil-elmoussaoui:
No idea :p

palob:
Could you check if it is the ``X`` in window headerbars?
If it is the symlink will introduce visual inconsistency there.

bil-elmoussaoui:
Nop not used at all! i got the icons from the source code

palob:
Yeah, you're right. It isn't included in Adwaita.
And as far as I can see the Numix ``window-close-symbolic`` looks the same as the Adwaita one.

PR #829, opened 2015-10-30
Highcontrast

bil-elmoussaoui:
Add symlinks from #827 
- user.svg
- folder-saved-search-alt.svg
- alarm.svg
- call-end.svg
- color-select.svg
- font-select.svg
- star-new.svg
- tool-check-spelling.svg
- headphones.svg

PR #900, opened 2015-11-22
Emotes icons

wa4557:
Introducing a new folder `emotes`. They are on #580 #829 and #437 . Scalable icons I will contribute in a subsequent PR


palob:
They can be ticked off from #577 too.

Foggalong:
Great work on these :+1:

PR #908, opened 2015-12-06
fixes #904 (application/vnd.geogebra.file)

### LINKED ISSUE #904: Geogebra Mimetype
Name: application/vnd.geogebra.file
Use: Geogebra graph files
Icon: 
screenshot_2015-12-05_20-41-22

hlechner:
fixes #904 
Added mimetype: application/vnd.geogebra.file

### DIFF
--- a/Numix/16/mimetypes/application-vnd.geogebra.file.svg
+++ b/Numix/16/mimetypes/application-vnd.geogebra.file.svg
@@ -56,7 +56,7 @@
      id="namedview12"
      showgrid="true"
      inkscape:zoom="64"
-     inkscape:cx="6.4837105"
+     inkscape:cx="7.6263787"
      inkscape:cy="6.3401012"
      inkscape:window-x="0"
      inkscape:window-y="27"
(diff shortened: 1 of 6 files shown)


palob:
Sweet :+1:

Foggalong:
The 22/24 symbols feel like they need to be shifted slightly to the right to counteract the triangle illusion

hlechner:
one pixel is enough?
image

palob:
I think that looks worse than before.

hlechner:
I tried to put it less than 1 pixel to right, but it was so blurry.

hlechner:
I checked that the whole symbol was 10.5 of width, so I change it to 10:
image

So it is better centered now, what do you think?

Foggalong:
Yeah, 1px definitely too much

Foggalong:
Same again. y u use stokes?

hlechner:
I didn't know at that time about strokes, I took the geogebra svg icon and I modified it, however the strokes was on it, I'll fix them.

palob:
Yeah, a few old Circle icons still use strokes.

Foggalong:
These icons are still using strokes and invisible shapes all over the place :(

hlechner:
Holy sh**, I probably was miles away. I don't know why I fixed only the small circles but not the rest...

but anyway, I think I found a way to discover if there is any stroke or shapes:
```
```

if someone with experience knows a better way to do it, please let me know.

I'll later check in my other Pull Requests for this too.

Issue #870, opened 2015-11-15
Octave missing icons


bil-elmoussaoui:
There's a few missing icons in Octave
![capture du 2015-11-15 11-45-05](https://cloud.githubusercontent.com/assets/7660997/11168415/64713936-8b8e-11e5-8e9b-6fce392be5e1.png)
The icons names are from left to right
- [x] bp-toggle -> media-playback-record
- [x] bp-prev
- [x] bp-next
- [x] bp-rm-all
- [x] db-step
- [x] db-step-in
- [x] db-step-out
- [x] db-cont -> media-playback-start
- [x] db-stop -> media-playback-stop

palob:
@bil-elmoussaoui Do you face blurry Numix icons in Octave?
![octave](https://cloud.githubusercontent.com/assets/7050624/11284326/6abecf00-8f09-11e5-9df0-21a5b23ce6f8.png)

bil-elmoussaoui:
@palob only when  you i choose small icons size in settings

palob:
Hmm, for me small icons result in the least blurry icons.

palob:
It's probably the 16px icons scaled up to 24px and 32px instead of using the 24px and 32px ones.

palob:
I'm working on those. 16px ``bp-*`` icons:
![bp-icons](https://cloud.githubusercontent.com/assets/7050624/11613847/23693260-9c30-11e5-8103-71baf73d2aee.png)

bil-elmoussaoui:
Nice!! :D

palob:
``db-*`` icons in 16px:
![db-icons](https://cloud.githubusercontent.com/assets/7050624/11614049/59218492-9c36-11e5-9a69-5437736766fb.png)

wa4557:
Could we use the same design for the `distribute-*` icons (from #437, inkscape icons )

palob:
Well, looking at the ``distribute-*`` icons in ``elementary`` and ``breeze`` I think a distinct design is justified.

wa4557:
Sure but the general design idea could be similar :)

bil-elmoussaoui:
Octave will look so Numixy! :D

PR #909, opened 2015-12-06
Breakpoint action icons for Octave

palob:
All ``bp-*`` icons from #870, preview over there.

### DIFF
--- a/Numix/16/actions/bp-next.svg
+++ b/Numix/16/actions/bp-next.svg
@@ -7,15 +7,13 @@
    xmlns="http://www.w3.org/2000/svg"
    xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
    xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   width="16"
    viewBox="0 0 16 16"
-   height="16"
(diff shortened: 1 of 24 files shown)


Foggalong:
While I get what you were going for with the design in keeping it consistent with octave it doesn't quite work. Possibly try putting the symbols inside the circles? Or without the circles (not sure in what context they're used, I only ever use CLI octave)? Or another design entirely?

palob:
You can apply breakpoints to your code in order to only execute a specific part of the file.
The icons are for jumping from one breakpoint to another or removing all of them.

I think symbols inside the circle will work, just the brush is a bit problematic.
A cross or something inside instead would be a double destructive methaphor though.

wa4557:
I hink the red circle should stay, in MATLAB it is also a red circle...

bil-elmoussaoui:
The same in Code::Blocks

Foggalong:
Good with the red circles then, but yeah maybe try a different idea for the symbols inside them :)

palob:
Next iteration:
bp-icons2

Foggalong:
Much much better :+1:

palob:
Revised designs pushed :+1:

Issue #898, opened 2015-11-21
Mimetypes: Microsoft files


hlechner:
# Microsoft common files:

Help Files:
- [x] **CHM document** (application/vnd.ms-htmlhelp) [[INFO]](http://fileformats.archiveteam.org/wiki/CHM)
- [x] **WinHelp help file** (application/winhlp) [[INFO]](http://fileformats.archiveteam.org/wiki/HLP_%28WinHelp%29)

Cursor Files:
- [x] **Windows cursor** (image/x-win-bitmap) [[INFO]](http://fileformats.archiveteam.org/wiki/CUR)
- [x] **Windows animated cursor** (application/x-navi-animation) [[INFO]](http://fileformats.archiveteam.org/wiki/Windows_Animated_Cursor)

Other:
- [x] **RIFF container** (application/x-riff) [[INFO]](http://fileformats.archiveteam.org/wiki/RIFF)
- [x] **Internet shortcut** (application/x-mswinurl) [[INFO]](http://fileformats.archiveteam.org/wiki/Internet_Shortcut) - also reported on #580

---
## Help Files:

**Description**:  **CHM** and **HLP** are used to show help information on Windows platform.
![image](https://cloud.githubusercontent.com/assets/11244067/11320197/a8a3dfe4-9077-11e5-9845-19c87c7ba699.png)
- **CHM**: `standard::icon: application-vnd.ms-htmlhelp, x-office-document`
- **HLP**: `standard::icon: application-winhlp, application-x-generic`

**Design Association**: Help, Information, Manpage
**Note**: Could be used this mimetype for  **Manpage manual document** file.
## Cursor Files:

**Description**: **CUR** file is a static cursor and **ANI** animated cursor for Windows systems. nautilus used to preview (thumbnail) the **CUR** as a normal image.
![image](https://cloud.githubusercontent.com/assets/11244067/11320227/0d7a7c9c-9079-11e5-914f-29d193fc0de4.png)
- **CUR**: `standard::icon: image-x-win-bitmap, image-x-generic`
- **ANI**: `standard::icon: application-x-navi-animation, application-x-generic`

**Design Association**: Cursor, pointer
**Note**: Could be used this mimetype for **X11 Cursor** file. The image BMP use the following mime: (image/bmp), the CUR uses  (image/x-win-bitmap) instead.
### RIFF

[...]

PR #914, opened 2015-12-06
added cursor mimetypes

hlechner:
imageimageimageimageimageimage
imageimageimage
imageimageimage

**Mimetypes:**
- image/x-xcursor
- application/x-navi-animation #898 
- image/x-win-bitmap #898 

Further details: [here](https://github.com/hlechner/hl-numix-mime/issues/1)


Foggalong:
The symbol should be `#f9f9f9ff` to fit with the guidelines :)

hlechner:
Could be the color ``#f9f9f9`` but with some transparency?

The file with ``#f9f9f9ff``:
image

hlechner:
Do you think that I should stay with the same background color?

here some tests of other background colors:
imageimageimageimageimage


and here the original background color with ``#f9f9f9ff`` plus 85% of opacity of cursor symbol:
image

Foggalong:
No transparency, sorry. I'm alright with the colour as is :)

hlechner:
Fixed.

btw, could you provide me the link to the numix base guideline? I want to take a look into it :thumbsup:

Foggalong:
I'm not aware that we actually have written guidelines for the base theme, they're just sort of derived by looking at the patterns at all of the icons within a certain category.

I'm gonna have to do a major rewrite of the website fairly soon so I'll possibly add some more details guidelines then :)

palob:
@hlechner On your 64px the vertical line of the **I**s look blurry. 
Try to move the **I** letters to the left/right or just the nodes effecting the position of the vertical line even if the resulting **I** will be slightly asymmetrical.

hlechner:
image

I've made some changes, what you think @palob?

palob:
That's crisp now :+1: 
But the lateral lines of the **I** now look a bit short in contrast to existing icons. Is it an illusion?

hlechner:
image

The middle still with the same original size (2px), the top and bottom I need to remove less than 1px for each side L/R to avoid blurry there. should i change something?

palob:
What results prolonging it by 0.5 (or 0.2, 0.7 ..) to each sides in?
Blurriness might not be that much of an issue in horizontal direction.

hlechner:
1px:
image
0.7px:
image
0.5px:
image
0.2px:
image

which you think is the better?

palob:
Hmm, 1,2 or 4. Not sure.
I'd say check an existing (not too blurry) 64px mimetype with an I at the end and choose what's comes closest.

hlechner:
Some that I've checked are about 0.2, so I choose 0.2

Foggalong:
The 16px icon needs bringing down half a pixel to counteract the triangle illusion. It's possible some of the other smaller sizes need this too, but not checked them has thoroughly

palob:
Are the cursor symbols pure white?

hlechner:
@palob no, I've used `#f9f9f9ff` instead as Foggalong's recommendation to follow the guideline.

hlechner:
Sanity check:
```
```
**OK**, nothing found

Foggalong:
There are a few minor niggles with some of the symbols at specific sizes but I'm not sure how well they could be addressed without ruining the overall shape of the cursor. I'll merge now but we might come back to this some time down the line :)

PR #916, opened 2015-12-07
RAW image mimetypes

hlechner:
imageimageimageimageimageimageimageimageimageimageimageimageimageimage

imageimageimageimageimageimageimageimageimageimageimageimageimageimage

imageimageimageimageimageimageimageimageimageimageimageimageimageimage
imageimageimage

RAW Images mimetypes

Fixes hlechner/hl-numix-mime#7

Further details: [here](https://github.com/hlechner/hl-numix-mime/issues/7)


palob:
I was about to do this one however my idea was to just use **RAW** on all of them, except for DNG and EXR

hlechner:
I didn't know about this EXR format before.

Yea, its not raw format but I can create a EXR mimetype with text for this one, should I add in this pull request or in a new one?

palob:
Thanks, better a new one. It's ``image-x-exr`` by the way.

Foggalong:
Is there any way we could get this more aligned to grid? It's very blurry as is.

hlechner:
I'll do my best on these icons.

I thought wasn't that bad. which sizes do you think is acceptable or maybe all sizes is so blurry?

Foggalong:
It's not terrible, but the smaller sizes especially could do with better alignment

hlechner:
image -> image

better or worst?

palob:
The lower one I think is an improvement :+1:

ChuckDaniels87:
I agree with @palob. :+1:

Foggalong:
Yeah :+1:

hlechner:
imageimageimage

palob:
What's the with of the symbol on the 16px one? Can you get it any larger without looking goofy?

hlechner:
The symbol's size is 6x6 let me try 7x7 then

palob:
Better use an even width/height. Otherwise you'll end up with a .5 X-position when centering.

hlechner:
8x8 on 16px:
image
* It's the same icon size of symbol on 24px icon

palob:
How looks the symbol from 22/24px on the 16px icon?

hlechner:
The symbol from 24px (8x8) on 16px:
image

It's better then 6x6 or it's too large?

palob:
It's easier to recognise as aperture at least. I think the size is tolerable.

hlechner:
And how about up 8x8 to 10x10 into the 22/24?

Here the test:
image

palob:
:+1:

hlechner:
32:
imageimageimageimageimageimageimageimageimageimageimageimageimageimage
22/24:
imageimage
16:
image

I've also repaired the 32px text, some letters were a bit blurry

hlechner:
Guys, what do you think now? and how about the 48 and 64?

palob:
I think it's all fine now. 
If you've got some extra time: I wonder if 16px could be slightly crisper (don' know if that's possible), 64px symbol looks like it could be a bit (2px to 4px) bigger.

hlechner:
actual:
image
tests:
imageimageimageimage

It's hard in this size.

hlechner:
64px:
image

The symbol above is 4px larger then the previously. I think it's better, should I change the icons to this size?

palob:
The new 64px is great.

The leftmost 16px test might be a tiny bit better than before but I'm not sure. Let' see what others think.

wa4557:
Yeah the leftmost is better imo

dirtydancing:
Wow, great work!

As to the 16px size: to me, the actual one looks better than the proposed leftmost one. The actual one to me is not really more blurry, and the actual one`s "thickness" is more consistent with the larger sizes. The proposed leftmost one is just fine as well, though :)

hlechner:
Sanity check:
```
```
**OK**, nothing found

palob:
@hlechner There is the ``image-x-hdr`` mimetype for HDR images (*Radiance format*).

While it's not RAW I think your new mimetpe is fitting.

hlechner:
I agree with you @palob, the mimetype is not the most appropriated because it's not the RAW file, however the HDR is closer to RAW then a normal image at my perception.

I'll do it and then I send a new PR.

wa4557:
@hlechner can you wait with sending a new PR until the maintenance PRs have been merged? Otherwise there will be conflicts. Adding something new is good to go, but changing something will lead to conflicts...

hlechner:
@wa4557 ok, no problem

palob:
@hlechner Did you already work on the HDR mimetypes? Otherwise I can do it.

hlechner:
I'm involved in other project right now,  so i didn't finish it.

please go ahead if you can, thanks.

palob:
Alright :+1:

PR #917, opened 2015-12-07
Exr

hlechner:
imageimageimage

``image-x-exr``


palob:
Thanks!
I think the preferred way to this is to have symlinks to the default image mimetype for 16/22/24px as well even if is likely to default to that one anyway.

The letters on the 32px one look a bit slim. Is it bold font weight?

hlechner:
image

First I though it was just my impression, but it wasn't I probably miss clicked the bold or the apply button before. Also created the links.

wa4557:
One thing: You should convert the fonts to paths, because on some distributions the Ubuntu font might not be installed which results in errors...

hlechner:
I did it, and also repair manually each letter for a better appearance.

wa4557:
Oh sorry, I was just looking at the source code and there are still some remnants from the font left :)

hlechner:
Thanks I see now, it's strange because when I open with inkscape it shows as path.

the method that i used was: select the font and go to "Object to Path"

palob:
Hmm, that should do the trick actually.

hlechner:
new tests: I've manually edited the code and remove the part of ubuntu mention and it look exactly as before.

I selected the whole filename actually

hlechner:
on that lines if you replace by this:
```
style="fill:#ffffff;"
```

it will work as the same, just need the color otherwise will be black.

I think inkscape do not remove some kind of information from file even if it will not be used it anymore.

PR #919, opened 2015-12-07
Adds missing symbolic action icons

wa4557:
I've added all (almost) the missing scalable action from #437 and #577. I also changed a few icons (grid aligning, changed to numix design). Apart from `pan-*` and `selection-` icons I tried to come up with new designs


palob:
Nothing you've touched but how about a looking glass icon for ``edit-find-symbolic`` like in Adwaita?
Would make it look more in line with ``edit-find-and-replace-symbolic`` and more distinct from ``find-location-symbolic``.

Likewise ``view-fullscreen-symbolic`` isn't in line with the non-symbolic one (as opposed to the Adwaita one).

Moreover I stumbled upon ``object-inverse`` in ``scalabe``. I suppose that one needs a ``-symbolic`` suffix.

wa4557:
Oh yeah. A lot of the scalable icons might need a doover for sure. I think however that `edit-find-symbolic` is one of the original icons designed for Numix. All (or most) of the icons are also not aligned to grid and the page size is usually something like 16x16.0123, so I will try and go through them and change that...

palob:
``application-menu-symbolic`` and ``open-menu-symbolic`` probably is another inconsequent pair.

``view-compact-symbolic`` could use a design distinct from ``view-grid-symbolic``. Both are used by Cinnamon.

Anyway, not to confuse @Foggalong , the PR is ready ready to merge :+1:

PR #929, opened 2015-12-09
Missing symlinks/icons from elementary

wa4557:
Some icons from elementary #437. I changed also the colour from `tag-new.svg` because it was completely white which causes problems for light themes


wa4557:
@bil-elmoussaoui @Foggalong Can you tick of `notification-new-symbolic` and `audio-volume-muted-blocking-symbolic` from the list. I really like seeing how the number goes down:)

Foggalong:
Sorry, must have missed that one. Done :+1:

PR #931, opened 2015-12-09
Shotwell action icons

palob:
Regrettably most of Shotwell's action icons are hardcoded.

The two icons with size suffix are only present for those particular sizes.

### DIFF
--- a/Numix/16/actions/shotwell-16.svg
+++ b/Numix/16/actions/shotwell-16.svg
@@ -1,0 +1,1 @@
+image-gallery.svg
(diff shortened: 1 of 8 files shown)


Foggalong:
Is shotwell the one developed now by the elementary devs?

palob:
They picked up Geary recently, Shotwell was forked as Pantheon Photos last year but except for a headerbar GUI hasn't evolved that rigourosly since.

Foggalong:
Ah cool :)

PR #932, opened 2015-12-09
Distinct design for window-new

palob:
The old design is now ``view-right-new`` in line with ``view-right-close``.

``window-close`` now looks distinct from ``view-right-*``.

The 19th change is an alignment fix.


Foggalong:
:+1:

PR #933, opened 2015-12-10
gramps mimetype - fixes #319

### LINKED ISSUE #319: Mimetype: application/x-gramps-package
Name: **Gramps package**
Mimetype: **application/x-gramps-package**
Possible extensions: **.gpkg**

Description: **Export of family tree of [GRAMPS](https://gramps-project.org/) software**

image

Target file name: application-x-gramps-package.svg

hlechner:
imageimageimageimageimage

**Alternatives:**
48:
image
24:
imageimage
16:
image

fixes #319


palob:
I'd go for the 2nd 22/24 alternative.

Foggalong:
Aye, definitely 2nd alt for 22/24

Foggalong:
These icons seem to be constructed with strokes rather than paths. They need switching and then properly aligning to grid and then good to merge :)

hlechner:
done, I've changed the strokes into paths.

However it still with the same alignment because I did it before, so the sizes still the same, or do you mean I move the symbol up/down?

palob:
Just in case the stroke width wasn't exactly 1px before or the strokes werent positioned at x.5 px.

Are all nodes on whole number pixel positions now?

hlechner:
When I created it I've adjusted the size of each square/stroke, then now I changed it to path by "Strokes to path" option and also I've checked that the sizes still the same.

Foggalong:
Still a few paper cuts on this. Makes sure all your nodes along horizontal/vertical lines are on point and aligned to grid. Also, if you have shapes over lapping to fill in a hole just merge the paths, it reduces the file size

hlechner:
I need to be honest, the things like Union/Difference/Intersection/etc are for me like the dark side of inkscape lol, because I never make it works as i planned. I'll try, if i could not make it works I'll ask you guys help.

hlechner:
after some boring time i finished, could you check it please?

it seems that is ok now.

Foggalong:
Will do :+1:

Foggalong:
Much better :) Just move all the left hand side symbol nodes 1px to the left in 32, 48, & 64 so that the symbol is properly centred.

hlechner:
done

hlechner:
Sanity check:
```
```
**OK**, nothing found

PR #934, opened 2015-12-10
Changes design of gsm-3g* and adds scalable icons

wa4557:
Makes design of gsm-3g icons consistent with new wifi icons and adds the same icons as `network-cellular-signal-*` from #437 in `scalable`


wa4557:
Btw: I think this icon would be also a better template for the `nm-tech-*` icons I recently added. @Foggalong what do you think? (The letters need to stay however)

wa4557:
This is what I would envision the new icons should look like: 
unbenannt

Foggalong:
Woah, that's looks really really neat. Love it :heart:

wa4557:
Fine I'll change that in the next few days. I'll open another PR though, so this one can be merged :)

PR #938, opened 2015-12-11
TCPDUMP mimetype

### LINKED ISSUE #379: Mimetype: application/vnd.tcpdump.pcap
Name: **Network Packet Capture**
Mimetype: **application/vnd.tcpdump.pcap**
Possible extensions: **.cap** **.pcap**

Description: **Packet capture from network, produced by softwares like tcpdump**
image

hlechner:
imageimageimageimageimage

Fixes #379

### DIFF
--- a/Numix/16/mimetypes/application-vnd.tcpdump.pcap.svg
+++ b/Numix/16/mimetypes/application-vnd.tcpdump.pcap.svg
@@ -13,7 +13,7 @@
    id="svg2"
    version="1.1"
    inkscape:version="0.91 r13725"
-   sodipodi:docname="application-tcpdump_16-3-.svg">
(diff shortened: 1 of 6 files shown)


Foggalong:
This is looking a lot better, but use path-difference rather than overlapping circles to create the hole effect :)

hlechner:
Nice, now i think i understood how to do it.

I've converted the whole symbol in only one path, or should I make it in 2 paths?

Foggalong:
Using multiple paths is fine, it's just not a good idea if you're using paths to conceal parts of other paths like here

hlechner:
Sanity check:
```
```
**OK**, nothing found

PR #951, opened 2015-12-17
Type=Fixed for 16,22,24,32px icons

### LINKED ISSUE #948: Blurry 24px action icons in Qt/GTK2 filechoosers and Octave, VLC
blurry1
blurry2
blurry3

The issue is due to 16px 22px or 32px (don't know exactly) action icons showing up in place of 24px ones.

This issue isn't present with other themes like Adwaita, breeze or elementary.

Adwaita and breeze have set ``Type=Fixed`` for actions icons in all sizes.
elementary has set ``Type=Scalabe`` like Numix (and the same Min/Max sizes except for 24px) but they haven't got 22px action icons.

Setting ``Type=Fixed`` in the Numix ``index.theme`` for 16,22,24,32 actions and the respective @x2 ones (haven't pinpointed it any more specifically) fixes the issue.

What do we need ``Type=Scalable`` for anyway?
Are they any non-hardcoded action icons with other sizes than 16,22,24,32....?
Those would show up blurry anyway.

palob:
Changed the icon type to ``Fixed`` for all sub-48px icons except for ``16/emblems``.
Making the action icons fixed size fixes #948.

### DIFF
--- a/Numix-Light/index.theme
+++ b/Numix-Light/index.theme
@@ -11,63 +11,47 @@ Directories=16/animations,16/apps,16/panel,16/status,22/animations,22/apps,22/pa
 
 [16/animations]
 Size=16
-MinSize=8
-MaxSize=21
 Context=Animations
-Type=Scalable
+Type=Fixed
 
 [16/apps]
 Size=16
-MinSize=8
-MaxSize=21
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [16/panel]
 Size=16
-MinSize=8
-MaxSize=21
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [16/status]
 Size=16
-MinSize=8
-MaxSize=21
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [16@2x/animations]
 Size=16
 Scale=2
-MinSize=8
-MaxSize=21
 Context=Animations
-Type=Scalable
+Type=Fixed
 
 [16@2x/apps]
 Size=16
 Scale=2
-MinSize=8
-MaxSize=21
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [16@2x/panel]
 Size=16
 Scale=2
-MinSize=8
-MaxSize=21
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [16@2x/status]
 Size=16
 Scale=2
-MinSize=8
-MaxSize=21
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [22/animations]
 Size=22
@@ -78,116 +62,86 @@ Type=Scalable
 
 [22/apps]
 Size=22
-MinSize=8
-MaxSize=23
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [22/panel]
 Size=22
-MinSize=8
-MaxSize=23
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [22/status]
 Size=22
-MinSize=8
-MaxSize=23
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [22@2x/animations]
 Size=22
 Scale=2
-MinSize=8
-MaxSize=23
 Context=Animations
-Type=Scalable
+Type=Fixed
 
 [22@2x/apps]
 Size=22
 Scale=2
-MinSize=8
-MaxSize=23
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [22@2x/panel]
 Size=22
 Scale=2
-MinSize=8
-MaxSize=23
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [22@2x/status]
 Size=22
 Scale=2
-MinSize=8
-MaxSize=23
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [24/animations]
 Size=24
-MinSize=8
-MaxSize=31
 Context=Animations
-Type=Scalable
+Type=Fixed
 
 [24/apps]
 Size=24
-MinSize=8
-MaxSize=31
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [24/panel]
 Size=24
-MinSize=8
-MaxSize=31
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [24/status]
 Size=24
-MinSize=8
-MaxSize=31
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [24@2x/animations]
 Size=24
 Scale=2
-MinSize=8
-MaxSize=31
 Context=Animations
-Type=Scalable
+Type=Fixed
 
 [24@2x/apps]
 Size=24
 Scale=2
-MinSize=8
-MaxSize=31
 Context=Applications
-Type=Scalable
+Type=Fixed
 
 [24@2x/panel]
 Size=24
 Scale=2
-MinSize=8
-MaxSize=31
 Context=Panel
-Type=Scalable
+Type=Fixed
 
 [24@2x/status]
 Size=24
 Scale=2
-MinSize=8
-MaxSize=31
 Context=Status
-Type=Scalable
+Type=Fixed
 
 [scalable/actions]
 Size=16
--- a/Numix-Light/index.theme
+++ b/Numix-Light/index.theme
@@ -171,6 +171,13 @@ MaxSize=512
 Context=Emblems
 Type=Scalable
 
+[scalable/emotes]
+Size=16
+MinSize=16
+MaxSize=512
+Context=Emotes
+Type=Scalable
+
 [scalable/mimetypes]
 Size=16
 MinSize=16
--- a/Numix/index.theme
+++ b/Numix/index.theme
@@ -928,6 +928,13 @@ MaxSize=512
 Context=Emblems
 Type=Scalable
 
+[scalable/emotes]
+Size=16
+MinSize=16
+MaxSize=512
+Context=Emotes
+Type=Scalable
+
 [scalable/mimetypes]
 Size=16
 MinSize=16
--- a/Numix/index.theme
+++ b/Numix/index.theme
@@ -12,37 +12,27 @@ Directories=16/actions,16/animations,16/apps,16/categories,16/devices,16/emblems
 [16/actions]
 Size=16
 Context=Actions
-MinSize=8
-MaxSize=21
-Type=Scalable
+Type=Fixed
 
 [16/animations]
 Size=16
 Context=Animations
-MinSize=8
-MaxSize=21
-Type=Scalable
+Type=Fixed
 
 [16/apps]
 Size=16
 Context=Applications
-MinSize=8
-MaxSize=21
-Type=Scalable
+Type=Fixed
 
 [16/categories]
 Size=16
 Context=Categories
-MinSize=8
-MaxSize=21
-Type=Scalable
+Type=Fixed
 
 [16/devices]
 Size=16
 Context=Devices
-MinSize=8
-MaxSize=21
-Type=Scalable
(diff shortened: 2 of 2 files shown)


wa4557:
You could actually make all the larger places folders also fixed (except 256 maybe for bigger sizes) . But it doesn't matter actually

wa4557:
@palob Can you also add folders `scalable/emotes` in this PR. Otherwise I will have to open another PR that is conflicting with this one (I recently added two icons there...)

palob:
@wa4557 Which icons are you referring to?
Should I add
```
[scalable/emotes]
Size=16
MinSize=16
MaxSize=512
Context=Emotes
Type=Scalable
```
to the ``index.theme`` or do the folder ``scalable/emotes`` as symlink to a fixed-size emotes folder?

wa4557:
First one. I'm talking about this folder https://github.com/numixproject/numix-icon-theme/tree/master/Numix/scalable/emotes

palob:
Ahh yes, looks like I haven't pulled the latest git changes.

palob:
Done :+1:

wa4557:
Thx

wa4557:
@palob In case you haven't seen my last edit in the previous post, you have to add the same folder in Numix-Light as well :)

palob:
Aye :)

hlechner:
And how about the 8px icons that were created as 16px? in this case the fixed type couldn't be problematic?

I didn't tried it because I don't remember which icons actually were 8px.

palob:
Do you know of any place where 8px icons are used?

wa4557:
I think we shouldn't really care about the stupidity of having `8px` icons. It's relatively useless since screen resolutions become larger and larger... `16px` is really the smallest I have ever seen on my computer.

That being said, keeping the `scalable` part for the `16px` icons might induce exactly those errors why we removed them in the first place...

Foggalong:
Cinnamon use[s,d] 8px icons in some parts of the interface, but other than that they're basically nowhere. It doesn't really make sense to cover them in Numix - it's a lot of work for a vanishing use case

hlechner:
got it, make sense, too much work for a little result

Foggalong:
Now, regarding the pull request - didn't we have a big thing just a few weeks ago where we switched everything to scalable?

wa4557:
Yeah, but this is actually only necessary for sizes `48` and `64` because they are used for `96,128, 256 ...` the scalable part for the smaller sizes seem to introduce problems with Qt applications.... Numix is a complete theme so the scalable part is not really needed

Foggalong:
And does having scalable at 48 and 64 not cause those problems also?

wa4557:
The problem is that icons from larger sizes are used for smaller icons. So no, because there are no bigger sizes

Foggalong:
The larger sizes do technically exist through their creation in `index.theme` even though the files don't physically exist

wa4557:
Yeah that's however true. I'm not sure if there are additional problems; never thought much about the Qt way of selecting icons...

Foggalong:
What would be the consequences of going back to everything (except `/scalable` ofc) as `Type=Fixed`?

wa4557:
We would have to add the larger sizes again...

wa4557:
Never programmed Qt before but I think I have a somehow similar program that simulates the Qt way of selecting icons. For me it seems that only the smaller icons are blurry....

Code (@Foggalong you might add that to numix-labs as well?)
```
from PyQt5 import *
from PyQt5.Qt import *
from PyQt5.QtWidgets import *
import sys

class MyMainWindow(QMainWindow):

    def __init__(self, parent=None):

        super(MyMainWindow, self).__init__(parent)
        self.form_widget = FormWidget(self) 
        self.setCentralWidget(self.form_widget) 

class FormWidget(QWidget):

    def __init__(self, parent):        
        super(FormWidget, self).__init__(parent)
        self.layout = QHBoxLayout(self)

        sizes=[16,22,24,32,48,64,96,128,256]
        icon = QIcon.fromTheme('distribute-horizontal-right')
        for i, size in enumerate(sizes):
            label = QLabel()
            label.setPixmap(icon.pixmap(size))
            self.layout.addWidget(label)
        self.setLayout(self.layout)

app = QApplication([])
foo = MyMainWindow()
foo.show()
sys.exit(app.exec_())
```

and screenshot:
bildschirmfoto 2016-01-07 15 35 01

wa4557:
With @palob s branch
unbenannt

wa4557:
It is a bit hard to see the difference in the screenshots above, but if you run the code it is clearly visible...

wa4557:
Ah ok, looking at the screenshots indeed the bigger sizes also need the "Fixed" entry. Damn...

wa4557:
This improved code confirms what I was fearing: Qt doesn't seem to understand the `scalable` part in the index.theme

```
from PyQt5 import *
from PyQt5.Qt import *
from PyQt5.QtWidgets import *
import sys

class MyMainWindow(QMainWindow):

    def __init__(self, parent=None):

        super(MyMainWindow, self).__init__(parent)
        self.form_widget = FormWidget(self) 
        self.setCentralWidget(self.form_widget) 

class FormWidget(QWidget):
    def __init__(self, parent):        
        super(FormWidget, self).__init__(parent)
        self.layout = QHBoxLayout(self)
        sizes=[16,22,24,32,48,64,96,128,256]
        icon = QIcon.fromTheme('distribute-horizontal-right')
        print(icon.availableSizes())
        for i, size in enumerate(sizes):
            label = QLabel()
            label.setPixmap(icon.pixmap(size))
            self.layout.addWidget(label)
        self.setLayout(self.layout)

app = QApplication([])
foo = MyMainWindow()
foo.show()
sys.exit(app.exec_())
```

Foggalong:
@wa4557 Added the [second version](https://github.com/numixproject/numix-labs/blob/master/Andreas/qt-test.py)

wa4557:
Ok in conclusion to this, I think we can say, that @palobs branch is fine at least when looking at the above screenshots...

wa4557:
Ok forget this, there are several problems with the Qt lookup mechanism:

Current situation:
GTK/Qt
bildschirmfoto 2016-01-07 18 40 00

@palob branch
bildschirmfoto vom 2016-01-07 18-43-32

Foggalong:
oh damn, that's not good

wa4557:
No it isn't... One thing we can do: Change everything back to `Fixed`. 
Create folders `96` and `128` (and `96@2x` etc. oc) and symlink `48` to `96` (the actual size doesn't matter) and everything from `64` to `128`. 
In the `index.theme` we can make everything fixed except all the `128` folder which we use for all the bigger sizes...

palob:
@wa4557 Great example code!

So you mean renaming ``48`` to ``96``?

wa4557:
@palob You need both folders `48` and `96` but in the `96` folder it is ok to have icons which are `48px`. So symlinking from `48` to `96` (same for `64` and `128`) should be fine

palob:
Do we actually need the dedicated SVG designs in ``96``, ``128`` and ``256`` or is this just scaled up from ``64``?

wa4557:
The problem is, that we need a dedicated `96` folder because it cannot be scaled from `64px`

palob:
Are they distinct from the ``48`` ones?

wa4557:
No. Which means they can be symlinked

palob:
So all icons, including ``@x2`` will scale correctly regardless of ``Fixed?
And it doesn't break the GTK situation?

wa4557:
Oh yeah,didn't think about it is all fixed now. I'm afraid we have to add the bigger sizes then :(

palob:
What I wondered before, do non ``apps``/``places``/``mimetype icons really exist in above-``64px`` sizes, i.e is there app using such?

wa4557:
Nevermind my idea works :) All fixed (except `128` which still has the `scalable` part. Seems to be all good 
bildschirmfoto vom 2016-01-07 21-28-14

palob:
How about the HiDPI ones? ``Fixed or ``Scalable``?

wa4557:
Fixed. But the `Scale=2` has to stay...

palob:
And we better hadn't toss the ``96`` and ``128`` SVGs?

If you set ``scalable/apps`` to ``Fixed`` in the Numix Circle index the app icons (>48px) show up blurry in the Shell overview.

So will there be all fine with >64px mimetype icons?

wa4557:
Yeah because if you set `scalable/apps` to `Fixed` with `Size=16` the system is only showing `16px` icons. But if you have `48px` icons in the `96` folder and set `Size=96` `96px` icons are shown. 

I wouldn't toss sizes > 64px. Screens become bigger and bigger these days...

palob:
I meant tossing the non-symlink SVGs.

wa4557:
No I don't think we can toss them. The `places` icons are very distinct...

palob:
Alright, I'll add the symlinks and the respective sections to the index file and set everything to ``Fixed`` save for ``128`` and ``128@x2``.
``Scalable`` too for the ``256`` stuff?

wa4557:
Yeah all the `256` scalable. All the `128` also `scalable` except for `places` which needs to be fixed.

wa4557:
You can symlink whole folders from `48` to `96` (same for `64`) except for `actions` I guess. We have to symlink all icons there individually...

palob:
Why is that?

wa4557:
Because there exist already `action` folders in `96` and `128` so we can't symlink the whole folder... But a     `find . -iname "*.svg" -exec ln -s {} ../../96/actions \; ` should do the trick

palob:
I would have symlinked the whole folder and pasted the existing back in.

wa4557:
That wouldn't work, because then you would replace the icons in the source folder...

palob:
Well, ``find`` outputs the filenames with ``./`` at the beginning. Since I couldn't ``sed`` or ``cut`` that away in combination with ``find -exec`` I've ended up with this:
``SVGS=($(ls)); for i in ${SVGS[@]}; do ln -s ../../64/actions/$i ../../128/actions/$i; done``

wa4557:
Sorry for the confusion, but you can actually ditch the `128` symlinks if you make `96` Fixed and keep `64` as scalable. So only symlinks from `48`to `96`necessary

wa4557:
Ok I had a closer look at `places` (and the `places` related `action` icons) in `96` `128` and `256`; and the good news is they are NOT distinct. They are in fact only scaled up versions of the icons in `48` and `64`. This means that they can be removed, and the same procedure as for the other folders can be used. So my proposal is now

1. Remove folders `places` and `actions` from `128`, `96` and `256` (also in `numix-folders`, saves a lot of overhead)
1. Symlink all folders in `48` (except `notifications`) to a newly created folder `96`
2. Change everything to `fixed` in the `index.theme` except all the `64` folders (with `Minsize=8` and `maxsize=512`)
3. Symlink all folders to their respective `@2x` folders for HIDPI support (also with the same `Fixed` `scalable` procedure)

This should then work with both GTK and QT and also removes a lot of overhead from `folders` and makes it easier to add new styles there. 
@Foggalong @palob What do you guys think?

wa4557:
I've made a test repo to show how I envisioned the whole thing (a few modifications to my above list were necessary, QT really has problems when it comes to `Type=Scalable` but now it should work :)
https://github.com/wa4557/numix-icon-theme/tree/qt-test

Proof that it works:
unbenannt

palob:
@wa4557  Alright, if it works out smoothly go ahead. I'll close this PR.
I can do a last real-life test with Octave 4 for your branch later.

wa4557:
:+1: I've opened a PR: #990. I guess this one can be closed now. @palob What I wonder is if the `fileopen` icon in Octave now works

palob:
I'll take it a ride tomorrow.

PR #954, opened 2015-12-18
LIB mime - fix the icons to guideline

### LINKED ISSUE #559: Mimetype: application/x-shared-library-la
Name: **libtool shared library**
Mimetype: **application/x-shared-library-la**
Possible extensions: **.la**

Description: **Library file**. could use the same mimetype icon as #557 

image

hlechner:
imageimageimageimageimageimage

Fixes #559

### DIFF
--- a/Numix/16/mimetypes/application-x-sharedlib.svg
+++ b/Numix/16/mimetypes/application-x-sharedlib.svg
@@ -40,9 +40,9 @@
      inkscape:window-width="1920"
      inkscape:window-height="1013"
      id="namedview12"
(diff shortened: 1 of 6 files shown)


Foggalong:
@hlechner Try 888888 like the ISO mime add a bit more consistency

hlechner:
I've changed to ``#888888`` as requested.

I also checked that the 64px of ISO mime use ``#878787`` instead of ``#88888`` which is used in other sizes.

Do you want to change it? I can create a pull request.

Issue #826, opened 2015-10-30
Missing icons from Humanity


bil-elmoussaoui:
Well just to mention that there's some few missing icons from Humanity & Humanity-Dark

Humanity-Dark:
- [ ] audio-volume-muted-blocked-panel
- [ ] battery_third_fourth
- [ ] battery_two_thirds
- [ ] gsd-xrandr
- [ ] gtg-panel
- [ ] ibus-off
- [ ] stock_weather-sunny
- [ ] sunny
- [ ] system-file-manager-panel
- [ ] wifi-020
- [ ] wifi-040
- [ ] wifi-060
- [ ] wifi-080
- [ ] wifi-100

Humanity : 
- [ ] add-folder-to-archive
- [ ] application-install
- [ ] application-x-remote-connection
- [ ] application-x-vnc
- [ ] aptdaemon-delete
- [ ] aptdaemon-setup
- [ ] audio-player
- [ ] audio-volume-muted-blocked-panel
- [ ] battery_third_fourth
- [ ] battery_two_thirds
- [ ] cab_extract
- [ ] calibrate
- [ ] dialog-info
- [ ] edit-delete-mail
- [ ] emblem-cool
- [ ] emblem-OK
- [ ] force-exit
- [ ] gnome-control-centre
- [ ] gnome-fs-bookmark
- [ ] gnome-inhibit-applet
- [ ] gnome-mime-application-x-remote-connection
- [ ] gnome-mime-application-x-vnc
- [ ] gnome-panel-force-quit
- [ ] gnome-panel-notification-area
- [ ] gnome-session-halt
- [ ] gnome-session-hebirnate
- [ ] gnome-session-reboot
- [ ] gnome-session-sleep
- [ ] gnome-session-switch
- [ ] gnome-spinner-rest
- [ ] gpm-hibernate
- [ ] gpm-inhibit
- [ ] grandr
- [ ] gsd-xrandr
- [ ] gtg-panel
- [ ] gtk-dialog-warning-panel
- [ ] ibus-off
- [ ] invest-applet
- [ ] lpi-help
- [ ] lv_icon
- [ ] media-import-audio-cd
- [ ] media-player-48
- [ ] messagebox_info
- [ ] ogmrip
- [ ] palm-pilot
- [ ] plugin-notification
- [ ] preview-file
- [ ] process-idle
- [ ] reboot-notifier
- [ ] resapplet
- [ ] rip
- [ ] scim-setup
- [ ] sleep
[...]

PR #955, opened 2015-12-18
Three new icons for elementary

wa4557:
Two new icons and a symlink for #437

### DIFF
--- a/Numix/16/status/user-idle.svg
+++ b/Numix/16/status/user-idle.svg
@@ -1,0 +1,1 @@
+user-away.svg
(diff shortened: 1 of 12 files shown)


bil-elmoussaoui:
`user-idle`is in those lists too : #826 #577 #580

Foggalong:
In user-typing the circles seem to be a little bit off grid which makes for a slight amount more blur than there should be

2015-12-23-214137_668x729_scrot

wa4557:
No idea why this happened, anyways fixed now...

Issue #827, opened 2015-10-30
Missing icons from HighContrast


bil-elmoussaoui:
Those are the missing icons from HightContrast, they look really ugly with Numix! 

- [ ] abrt-gui
- [ ] action-unavailable
- [ ] alarm
- [ ] anjuta6
- [ ] applications-blender
- [ ] application-x-appliance
- [ ] bookmark-view
- [ ] call-end
- [ ] call-missed
- [ ] camera-video
- [ ] channel-insecure
- [ ] channel-secure
- [ ] checkbox-checked
- [ ] checkbox-mixed
- [ ] colorimeter-colorhug
- [ ] color-select
- [ ] computer-apple-ipad
- [ ] computer-fail
- [ ] config-firewall
- [ ] configuration-editor
- [ ] content-loading
- [ ] display-brightness
- [ ] display-projector
- [ ] documents
- [ ] drive-harddisk-solidstate
- [ ] drive-multidisk
- [ ] drive-multidisk-alt
- [ ] drive-multidisk-alt2
- [ ] edit-clear-all
- [ ] edit-select
- [ ] emblem-music
- [ ] face-angel
- [ ] face-angry
- [ ] face-confused
- [ ] face-cool
- [ ] face-crying
- [ ] face-devilish
- [ ] face-embarrassed
- [ ] face-glasses
- [ ] face-kiss
- [ ] face-laugh
- [ ] face-monkey
- [ ] face-plain
- [ ] face-raspberry
- [ ] face-sad
- [ ] face-shutmouth
- [ ] face-sick
- [ ] face-smile
- [ ] face-smile-big
- [ ] face-smirk
- [ ] face-surprise
- [ ] face-surprised
- [ ] face-tired
- [ ] face-uncertain
- [ ] face-wink
- [ ] face-worried
- [ ] face-yawn
- [ ] find-location
- [ ] folder-saved-search-alt
- [ ] font-select
- [ ] gtk-bold
- [ ] headphones
- [ ] headpin
- [ ] history-view
- [ ] ifolder
- [ ] input-dialpad
- [ ] internet-group-chat
- [ ] katello
- [ ] keyboard-brightness
- [ ] list-remove-all
- [ ] mark-location
- [ ] media-optical-bd
- [ ] media-optical-cd-audio
- [ ] media-playlist-consecutive
- [ ] media-playlist-consecutive-rtl
- [ ] media-playlist-repeat-rtl
- [ ] media-playlist-repeat-song
[...]

PR #956, opened 2015-12-20
Fixes wrong naming of network-cellular icons

wa4557:
I named the icons in `scalable` without the `symbolic` extension. I changed that and added the icons instead in the fixed size directories. Fixes some icons from #437


bil-elmoussaoui:
The new icons are in here #827 too

PR #967, opened 2015-12-23
Adds distribute-* icons

wa4557:
From #437 and #580.This is how they look like in inkscape 
bildschirmfoto 2015-12-23 15 19 45


palob:
Yay!

Foggalong:
\O/

PR #969, opened 2015-12-23
Bunch of symbolic icons for Adwaita

wa4557:
from #577


Foggalong:
We might have a Circle icon whos symbol could be used for `camera-video-symbolic.svg`

wa4557:
Was actually looking before I tried to make the icon but the only one close was `screen-recorder` and this one doesn't really qualify as a good icon...

Foggalong:
Thinking about it, I think of a lot of the "video" type apps use the film-tape symbols rather than camera symbols. Happy with this :+1:

wa4557:
We might revisit it as soon as `camera-video.svg` will be added...

Foggalong:
:+1:

PR #970, opened 2015-12-25
Enhancement for teamviewer indicator

wa4557:
The indicator was really big compared to already existing tray icons. Made the circle 1px smaller


Foggalong:
Looking much better :)

PR #971, opened 2015-12-25
A few symbolic icons for Adwaita

wa4557:
from #577. I've renamed (and also changed) one icon that was there `user-status-symbolic` to `user-status-pending-symbolic` because it was obviously wrongly named


wa4557:
I also added alignment fixes for three symbolic icons

Foggalong:
Good job :+1:

PR #977, opened 2015-12-31
Fixes #972

### LINKED ISSUE #972: Wrong icon for home, trash and desktop with Numix Light icons
screenshot-28-14 59 20

They all use the avatar icon

wa4557:
Fixes #927. This bug is related to #190. The icon `user.svg` in Numix-light caused problems because when a gtk application requests an icon that starts with `user` like `user-trash.svg` for example this icon is used instead of going down the inheritance line. 

Fortunately this icon does not need to be monochrome (because it's not a tray icon) so I changed it's color and only kept it in numix base, where it causes no problems

### DIFF
--- a/Numix-Light/16/status/user.svg
+++ b/Numix-Light/16/status/user.svg
@@ -1,1 +1,0 @@
-avatar-default.svg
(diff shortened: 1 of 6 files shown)


dirtydancing:
Confirmed that this fixes https://github.com/numixproject/numix-icon-theme/issues/972.

Foggalong:
:+1:

PR #978, opened 2015-12-31
Add missing goa-account icons

forivall:
screenshot from 2015-12-31 14-58-56
Added missing and symlinked broken goa-account icons. Continuation from numixproject/numix-icon-theme-circle#2754.

I aligned the icon sizes to the size of the google icon & owncloud icon & stock goa icons, rather than the size of the flickr and msn/live & pocket icons. Perhaps those icons could be made a little bigger?


bil-elmoussaoui:
@forivall Can you add Foursquare icon too? :)

forivall:
Sorry, I don't have the graphics design skills to do it quickly, and not enough time right now :dash:. Maybe later, if this PR hangs around a long time :wink:

wa4557:
I'm afraid you have to add all the sizes up to 64px like the already existing ones

dirtydancing:
And how about the icon for "Other": would that be `goa-account.svg`? "Other" shows the current `goa-account.svg` icon with Numix-Circle, just like currently all accounts are showing with Numix-Circle.

bil-elmoussaoui:
i see why the goa-account in Hicolor is transparent! We need to add it too..

wa4557:
So this then needs to be added in the `base` theme rather than circle?

bil-elmoussaoui:
Yes i will add it to #979

Foggalong:
@forivall as said above, these symlinks/icon need adding for all sizes not just 48px :)

forivall:
I added the other sizes. It might be useful to create a resizing script; I did it by surrounding the contents with a transform, and then running `for d in 16 22 24 32 64; (sed -e 's/viewBox="0 0 48 48"/viewBox="0 0 '"$d $d"'"/' -e 's/transform="scale(1)"/transform="scale('$(($d/48.0))')"/' 48/apps/goa-account-lastfm.svg > $d/apps/goa-account-lastfm.svg)`

The extraneous <g> can then be removed by svgo

Foggalong:
@forivall Resizing scripts generally don't work because things still have to be aligned to grid

forivall:
grumble grumble pixel grid.
I aligned the 22px icons, and otherwise, the other sizes were already aligned properly.

Foggalong:
The problem is more wide spread than that - the icon I opened to check before was 16px and that wasn't aligned, which also probably means 32px, and 64px aren't either.

forivall:
ah, it's the inner edges of the windows icon. my bad, haven't done this before.

Foggalong:
No problem, we all start somewhere! I'll review this again tomorrow :)

PR #982, opened 2016-01-01
Removes clipping from folders with fewer icons

wa4557:
Ok these are the folders where the cleaning procedure (removes clipping + svgclean script) can be applied without a conflict with current PRs. Only `mimetypes` (a bit more complicated because it seems that a few icons are indeed still using clipping) and status (also a bit complicated because there are a few text elements left) and all of the `scalable` icons now need this treatment


wa4557:
Btw: all this cleanup saves ~6MB of sizes (without the biggest part, which is status/places icons)

PR #984, opened 2016-01-03
Maintenance for mimetype icons

wa4557:
Since not there are no more mimetype related PRs, here is the same maintenance procedure for mimetype icons. Saves about 3MB...


Foggalong:
@wa4557 A commit involving status icons seems to have slipped in here

wa4557:
Rebased :+1:

PR #986, opened 2016-01-03
Maintenace for scalable icons

wa4557:
Same treatment for scalable icons. Together with the maintenance for the places icons and all the maintenance PRsThe filesize reduction is roughly 40MB. So quite a bit :)


Foggalong:
@wa4557 `Numix-Light/scalable/places/user-home-symbolic.svg` seems to have been borked

wa4557:
Fixed

wa4557:
@Foggalong Can you copy the icons from `numix-folders` to here as well? That's actually the part where most of the filesize reduction is possible...

Foggalong:
Sure thing :)

PR #990, opened 2016-01-08
Fixes qt icon lookup failure

wa4557:
Fixes Qt icon lookup breakage. What have I done:

1. Removed `places` and `actions` from sizes bigger than `64px` (only scaled up versions of `48px` and `64px`)
2. Symlink all folders from `48` to `96` (except `notifications`)
3. Symlink all folders from `64` to `128`
4. Completely removed `256`
5. Changed everything (except `scalable` folders o.c.) to `Type=Fixed` in the `index.theme` except
6. All folders `128` got `Type=Scalable` with `MinSize=97` and `MaxSize=512`

These new "virtual" folders ensure that the icons are taken from the right folders for GTK and Qt.
unbenannt

The same fix has to be applied oo `numix-folders` package


palob:
@wa4557  No more blurry icons in Octave and other Qt apps when running your branch :+1: 

From #948 only the edit icon in the GTK2 filechooser remains blurry. But it's blurry with even ``Adwaita``, looks like a GTK2 issue.

wa4557:
Ah the folders from `numix-folders' have created a merge conflict. Do hard to repair. I'll open a new PR

PR #994, opened 2016-01-19
Fixes qt icon lookup failure

wa4557:
See #990


palob:
@wa4557 Can you test this branch with Dolphin? 
In Dolphin you can change the icon size dynamically, wondering if that needs scalable icons.

wa4557:
@palob No problem there. The Qt lookup mechanism seems to be working different to the GTK one. While GTK does not interpolate between sizes (if Type=Fixed) Qt does, no matter what the "Type=" is

palob:
Merge conflict due to the Qt fix commit.

PR #995, opened 2016-01-19
Fixes for broken scalable icons

wa4557:
Changed strokes to paths and also made the document-open-symbolic icon better recognizable


palob:
Thanks for this one. I think ``document-open`` might be even more recognisable with a gap in the paper sheet outline like in the Adwaita icon:

adwaita-document-open

wa4557:
:+1:

PR #998, opened 2016-01-23
Fixes qt icon lookup failure

wa4557:
And another one. @Foggalong Can you take a look at this. I know there are a lot files, but there's not really a lot that is changed. Removing the `places` icons from `96`,`256` and `128` also makes future development easier (They are really nothing else than upscaled versions from `64` and `48`)


Foggalong:
I'll try and get this looked at asap :+1:

wa4557:
Proof that the `96` `128` and `256` are redundant
96px
test
48px
test

64px
test
128px
test
256px
test


And this is the same for all styles (6 is the most complicated anyways). So really this should be merged.

Foggalong:
Sorry for the delay on this, thought I'd merged it ages ago :flushed:

wa4557:
Don't worry, but you have to merge the corresponding folder PR as well :)

PR #1022, opened 2016-03-19
Adds object-group.svg

wa4557:
one from #437


wa4557:
I've added also the related `object-ungroup`

palob:
Oh, what an improvement compared to the default!

PR #1029, opened 2016-03-24
Preferences icons

wa4557:
three from #437. These icons already exist in circle


wa4557:
@Foggalong I based the design on the icons from your Technic icon theme (albeit I changed it slightly). I hope that's ok...

Issue #1020, opened 2016-03-13
Mint Update Icons


Foggalong:
Mint update icons are [no longer hardcoded](https://github.com/linuxmint/mintupdate/issues/133#issuecomment-196050264) so we need to cover them. Information required is found via link above.

bil-elmoussaoui:
This issue is the same as this one #999 ?

Foggalong:
Different icon names now :)

bil-elmoussaoui:
I will change the icon names in Hardcode-Tray too

Foggalong:
@bil-elmoussaoui Ah, do you not want to keep #999 open then?

bil-elmoussaoui:
Nop, just close it :+1: it's the same issue, once the icons are added for this one, the other one will be fixed too

wa4557:
Are thos icons shown in the tray?

bil-elmoussaoui:
@wa4557 Yes

wa4557:
I guess the `mintupdate-level1` and `mintupdate-type-backport` are not, right?

Foggalong:
@palob @wa4557 @bil-elmoussaoui 

The others are still status icons as far as I can tell, just used within the app to show the status of each package. Any Mint users out there feel free to correct me if I'm wrong though.

PR #1034, opened 2016-03-27
Adds mintupdate tray icons

wa4557:
First part of #1020. Not sure where the other icons are supposed to be


palob:
If you are unsure where to put them (``status``,``actions``,...) that doesn't actually matter from my experience. The icons will show up correctly anyway.

wa4557:
Yeah sure but I don't know which kind of icons they are supposed to be. Are they action or status icons? (There is a difference in designing them, right?)

Foggalong:
Afaik these are status icons used by the update software in the Cinnamon tray.

wa4557:
Yeah the ones in this PR, sure. But arguably not the other ones that are linked in #1020

PR #1040, opened 2016-03-31
Misnamed KML mimetype deletion

palob:
Correclty named one is in place.


Foggalong:
Are we sure this is misspelled rather than just a different distributions naming preference?

palob:
Hmm, I did install this one in the first place. Officially there isn't any ``vnd-`` mimetype. 
However it is in ``breeze`` and ``oxygen`` this way. Maybe we'd better play it safe.

PR #1051, opened 2016-04-04
Adds more inkscape icons

wa4557:
two more icons for inkscape (the distribution icons are still mixed)

all the inkscape icon names can be found here:
http://wiki.inkscape.org/wiki/index.php/Themable_icons


Foggalong:
Might be worth coming back to that document after we've done the base lists :+1:

palob:
@wa4557 Thanks for finding this list and your icons.

I think ``breeze`` got all those icons covered so this is part of #580 as well.

PR #1054, opened 2016-04-09
Adds model.svg

wa4557:
From #437 (only a few more icons left \o/)


wa4557:
Btw:  `application-x-fictionbook+xml` can also be ticked of from #437, that's already there

PR #1059, opened 2016-04-12
Adds library-audiobook.svg

wa4557:
from #437


Foggalong:
@wa4557 Do we know where this icon is used?

PR #1061, opened 2016-04-14
Adds library-places

wa4557:
from #437


palob:
That's a rad one :+1:

Foggalong:
Same with this icon - do we know where this icon is used?

wa4557:
No plan. I just thought that since it is not a folder icon something similar to the non folder icons in places might be the right choice

Foggalong:
That's my worry - if it's something representing a directory in the file system it needs to be a folder to stay consistent with the others. Anyone who uses elementary know where this shows up?

bil-elmoussaoui:
@Foggalong Those are the default icons from elementary-icon-theme
screenshot from 2016-06-20 00-47-24

Foggalong:
@bil-elmoussaoui Any idea where they're used though?

bil-elmoussaoui:
@Foggalong Never seen those icons anywhere. But there still in the latest version of elementary-icon-theme (launchpad); they only exists on 16px; i think they're from luna..

palob:
If they're only made for 16px I think it's quite probable they won't appear in place of folders.

Foggalong:
Probably safe to merge these two issues then and just see if they crop up somewhere they shouldn't :)

PR #1064, opened 2016-04-24
Adds new-design for indicator-keyboard

wa4557:
I've updated the keyboard-indicators to the "new" style, which means that I've changed the colours and also removed the shadow. Note that I scripted these changes, which means that I didn't change the design but only adapted them so that the keyboard indicator doesn't look alien in the tray anymore...


Foggalong:
That's fine, thank you so much for addressing this. I'd honestly forgot that it needed to be done :astonished:

wa4557:
Yeah, but that's not worse than it was before... Unfortunately those icons are designed using masks and `<text>` elements, and I haven't found a way to script this in order to use `paths`. 

Frankly 3700 icons are just too much in order to go through them by hand

wa4557:
I could probably script something like a negative of the current one, without the square and only the letter; not aligned to grid oc, but probably better than the current one...

Foggalong:
I'm not too bothered about aligning these all to grid for now, it's more that in some cases important text is being cut off the icons (some have 3 characters rather than just 2 and the third is often unreadable)

Foggalong:
:heart:

PR #1075, opened 2016-05-27
Mimetypes for Haxe, Nim, Crystal + Fixes for Ruby

TiZ-EX1:
I was thinking about using a darker color for Crystal since its main color seems to be straight up black, but the darkest color in the pack so far is the color it's already using. If you think it's alright to go darker, just tell me how dark you wanna go and I'll amend the PR.

Also, Ruby's icons had a bunch of inconsistencies so I fixed them.


palob:
Where did you got those mimetypes from? Are they registered by any app?
I think those aren't in the [shared mime info database](https://freedesktop.org/wiki/Software/shared-mime-info/) yet which means the mimetype icons won't show up.

Anyway, thanks for your contribution, we can add the icons for the future.
Your designs are fine, except for the Nim symbol placement.
+ For ``16`` it should be moved 0.5 px down
+ On the ``22`` icon it should be at ``Y=5`` and be 8 or 8.2 in height
+ ``24`` icon should use the same symbol like ``22`` at ``Y=6`` (1 px higher ``Y`` than ``22`` in general)
+ ``32``: ``Y=8``, height 12px
+ ``48``: ``Y=13``, height18px
+ ``64``: ``Y=16``, height 24px

I'm still wondering however whether the Nim symbol is sufficiently recognisable or a most simple 2D crown view (a bar with triangles on top) could work better.

TiZ-EX1:
Hi there, thanks for the change suggestions. I was just eyeballing everything for Nim, so it's good to have some precision suggestions. The node changes didn't show up in the email, so I'll go back and look at those. Am I correct in assuming the blue ones are the ones that can go?

I'm personally of the opinion that a 2d crown would be too generic to be recognizable.

As for where these mimetypes even came from... well, they're not officially supported anywhere. I get why Crystal and Nim aren't in the database what with how young they are, but I don't get why Haxe isn't, because it's definitely *not* young. Anyways, this is a case of me fixing it for my own use and contributing back because why not.

palob:
Yep, the blue ones can go.

Have you got any Crystal/Nim/Haxe source files at hand? Right-click on them in a file manager and open the file properties dialog. There you can see whether a mimetype is assigned or it is just treated as ``text/plain`` (in which case the mimetype icon won't show up).

TiZ-EX1:
Right now they have mimetypes because I made some custom ones in ~/.local/share/mime/packages. But I had to do that for all of them. `grep -r Haxe /usr/share/mime` shows nothing, same for Crystal and Nim. If I wanted to contribute my custom mimetype definitions, where and how should I do that?

palob:
Maybe submit a [patch](https://freedesktop.org/wiki/GettingInvolved/) (I haven't got the faintest idea what you'll have to do) or a [bug report](https://bugs.freedesktop.org/buglist.cgi?bug_status=__open__&order=relevance%20desc&product=shared-mime-info&query_format=specific) against the [shared mime info repo](https://freedesktop.org/wiki/Software/shared-mime-info/)?

It looks like all relevant specs are stored in the following XML files:
https://cgit.freedesktop.org/xdg/shared-mime-info/tree/freedesktop.org.xml.in
https://cgit.freedesktop.org/xdg/shared-mime-info/tree/shared-mime-info-spec.xml

palob:
@TiZ-EX1 In case you're interested in submitting the mimetypes to ``shared-mime-info`` [here's](https://github.com/probonopd/AppImageKit/issues/144) a discussion about a submission for another mimetype.

TiZ-EX1:
Thank you for the information about that.

PR #1086, opened 2016-06-11
Fixes #1050

### LINKED ISSUE #1050: Missing view-app-grid-symbolic for Gnome 3.20
Gnome have changed the icon name from `view-grid-symbolic` to `view-app-grid-symbolic`. 
More informations about that : https://plus.google.com/+WorldofGnomeOrg/posts/1ag97fAXQ67

wa4557:
#1050

### DIFF
--- a/Numix-Light/scalable/actions/view-app-grid-symbolic.svg
+++ b/Numix-Light/scalable/actions/view-app-grid-symbolic.svg
@@ -1,0 +1,1 @@
+view-grid-symbolic.svg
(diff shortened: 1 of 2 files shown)


palob:
The current Adwaita one has circles instead of squares like in ``view-grid-symbolic``.

I think this is a better match when using Numix-Circle (circular icons in the GNOME app overview).

Is it reasonable to ship the icon with Circle/Square?

wa4557:
So, you suggest keeping `view-grid-symbolic` and changing only `view-app-grid-symbolic`?

palob:
Yes, a distinct ``view-app-grid-symbolic`` design (circles instead of squares). Including different ones in Circle/Square however entails having action icons in those repos.

PR #1092, opened 2016-06-21
Bold font for indicator-keyboard icons

wa4557:
Alright, here we go. Looks indeed better, but there are probably a few icons where the letters are now bigger than the baseplate. I didn't look at all of them however and I tried to adjust spacing and font-size accordingly


wa4557:
@Foggalong I've gone through the icons by hand, and made sure that all the font is now within the baseplate

PR #1098, opened 2016-07-10
Alignment fixes for printer-symbolic

wa4557:
The icon is shown in the tray whenever something is getting printed. Lines were not aligned to grid


Foggalong:
@wa4557 The internal rectangle cut still needs to be there otherwise the design makes no distinction between the front and the top of the printer

wa4557:
Alright pushed. I just thought that too much details are not necessary in the tray. The old design was also a bit more distinct from the Adwaita ones, and looked nice in the tray (and more consistent with all the other icons imo)

PR #1099, opened 2016-07-10
Alignment fixes for owncloud status icons

wa4557:
Icons were definitely not the best looking before, so aligned them properly and adjusted the spacing. I also removed the checkmark from `state-ok` since this is consistent with all the other cloud based services, where only the unmodified logo is shown, if everything is ok (dropbox, copy, bittorrent-sync etc.)


Foggalong:
As much as I'd like to have the icons aligned to grid, the implementation in this merge doesn't work with the current ownCloud design. So much of the icon is being taken away in cutaways that it's not even recognisable as the cloud.

iirc ownCloud also makes a distinction between just running and having state-ok, though I don't have a machine with it on right now so I'll have to verify that's the case.

wa4557:
I can understand the too much cut away argument, although I think this is not a big issue. After all, these icons are only shown intermediately and only if something is going on, and changes to the generic owncloud icon eventually. It's just that the icons like they are now are not fitting with the rest of the theme imo (too little space between symbols and icon).

Concerning the `state-ok`: Hmm yeah but I think it is more important to be consistent with other icons in the theme, not with the original ones. We don't have to copy them. All the other cloud based icons are using the logo as `-ok` icon

Foggalong:
It only takes two icons in the panel to look like pauses before things get confusing and I'd much sooner avoid that where possible.

If I have got this right (I don't have a functioning ownCloud server to test it with at the moment) there's a literal functional distinction between the program running and the state of the program being okay, so having distinct icons isn't a realistic option.

wa4557:
Ok agreed on the `-paused` icons, didn't think about them. The `synchronizing` icon looks nice though. Also the `error` icon is easily recognizable (and the `information` icon is hardly ever used) 

I think I can do something about the `paused` icon. I think we agree that the icons right now are not properly aligned and designed (just look at the checkmark of `state-ok`; it looks completely weird)

Sorry, but no. I'm running an owncloud instance for the last 5 years, and I have never seen the `owncloud-status` icon so far in the tray...

Foggalong:
In which case I'm probably misremembering - I'm happy to go with the symlink :)

PR #1100, opened 2016-07-10
Fixes #989

### LINKED ISSUE #989: nm-applet cellular icon is broken
image

Reproduce:

1. Install Numix icon theme

wa4557:
Addresses #989. Still a pity that these icons are handled like this...


Foggalong:
Yeah :( do you know where the upstream report needs filing with? I don't believe the original reporter left the distribution

Regarding the redesign, I'm not super comfortable with the separation of emblems with 2 characters. I'd sooner have the two together and slightly smaller and the cost of having them slightly less readable.

wa4557:
No idea, could be a n issue with network-manager? 

Alright, I'll try something else, it's just that the wireless icons in Numix are bigger than for other themes, which means that there's less space.

wa4557:
I'll work on this on a new PR

Issue #1107, opened 2016-07-19
Electron apps tray icons


bil-elmoussaoui:
Here's a list of new tray icons that were added to Hardcode-Tray recently and i think that have a lot of users out there :)

Nylas : 
- [x] `nylas-tray-full-newitems`
- [x] `nylas-tray-full`
- [x] `nylas-tray-empty`

Stremio: 
- [x] `stremio-tray`

Twitch Indicator:
- [x] `twitch-indicator`

Anatine: 
- [x] `anatine-indicator`
- [x] `anatine-notification`

![selection_003](https://cloud.githubusercontent.com/assets/7660997/16900999/48ff73e6-4c2e-11e6-919f-5ede53b5352c.png)

Caprine: 
- [x] `messenger-indicator` 
- [x] `messenger-notification`

![selection_002](https://cloud.githubusercontent.com/assets/7660997/16900973/8092f1da-4c2d-11e6-8085-1a82e5a137a2.png)

Slack:
- [x] `slack-indicator-highlight`
- [x] `slack-indicator`
- [x] `slack-indicator-unread`

I'm going to add screenshot of icons later.

bil-elmoussaoui:
@Foggalong @wa4557 I think this one can be closed ? :)

wa4557:
yeah

PR #1109, opened 2016-07-23
Electron tray icons

wa4557:
A bunch of tray icons from #1107


Foggalong:
All good apart from the following:

* Replace `nylas-tray-newitems.svg` with `nylas-tray-full.svg`  and move `nylas-tray-empty.svg` to `nylas-tray-full.svg`. Then create a new icon which is a transparent version of `nylas-tray-full.svg` and name it  `nylas-tray-empty.svg`. This allows for distinction between new items and a full tray.
* What's the distinction between highlight and unread in Slack? I'm not familiar with that program

wa4557:
Check on the `nylas` icons, will push that asap

Messaging app. Numix does not differentiate between `unread` and `highlight` so I guess using the same icon is fine. I could use the warning orange though?!

Foggalong:
Nah, keep them both blue for Slack :)

wa4557:
Pushed

PR #1116, opened 2016-08-07
internet-radio and library-podcast

wa4557:
One new icon and a symlink


Foggalong:
I was going to protest about the use of the network icon for podcasts, but on thinking about it I realise how apt it is. Nice work :+1:

Issue #1110, opened 2016-07-25
Icon request: qtox tray icons


frisbee00:
The tray icon names for qtox as per https://github.com/tux3/qTox/pull/2689.
* [x] **qtox-online**
![qtox-online](https://cloud.githubusercontent.com/assets/6317854/17097735/aebd9a74-527c-11e6-86d2-9604db4b2e88.png)
* [x] **qtox-offline**
_Could not find the right icon._
* [x] **qtox-busy**
![qtox-busy](https://cloud.githubusercontent.com/assets/6317854/17097737/b88bc77e-527c-11e6-927d-7c5fdb36cc71.png)
* [x] **qtox-away**
![qtox-away](https://cloud.githubusercontent.com/assets/6317854/17097743/c1524b8a-527c-11e6-8138-fbb99bee9d4e.png)
* [x] **qtox-invisible**
_Could not find the right icon._
* [x] **qtox-event**
_Could not find the right icon._

bil-elmoussaoui:
The icons can be found here https://github.com/tux3/qTox/tree/master/img/taskbar/

frisbee00:
How about these for starters?
[qtox-panel.zip](https://github.com/numixproject/numix-icon-theme/files/399234/qtox-panel.zip)

bil-elmoussaoui:
You need to provide 16x16, 22x22 and 24x24 sizes also a dark and light variant of colors. Once it's done you can open a PR

frisbee00:
Are those fine? I am not sure about the colours and other specifications for it to fit in the theme.

bil-elmoussaoui:
I can't really give you an opinion about this. @wa4557 is the best one to comment your tray icons design! :+1:

Foggalong:
If these are to go in the tray they need to be monochrome or single colour for notification highlights. They also need aligning to grid considerably more before they can be merged. @frisbee00 if you'd like to work on them with us though open a pull request and we can start on fixing them up :)

wa4557:
Fixed \o/

PR #1117, opened 2016-08-09
Initial set of tray icons for qtox

frisbee00:
Added a set of light tray icons (24x24) for qtox.
The following are changes from the set posted in #1110 

- Monochrome versions
- Partial alignment with the grid

I'm not sure of the name `qtox-event-small` (tentative name based on names from https://github.com/tux3/qTox/tree/master/img/taskbar/ and https://github.com/qTox/qTox/pull/2689) and other specifications for the theme.
Other sizes and dark versions can be added once we have an acceptable design.


palob:
Can you work on the grid alignment for ``qtox-invisible`` for better recognisability?

For instance by subtracting the red rectangle from the path:
invisible

frisbee00:
Tray icons from qtox have no white space above or below. I tried to fit the icons to the Numix theme; used https://github.com/numixproject/numix-icon-theme/blob/master/Numix/24/panel/redshift-status-on.svg as a reference (4px above and below).
As for the circle diameter being an odd number, 6px appeared too small while I tested and 8px too big to fit inside without modifying the original tox icon considerably.
Will modify `qtox-invisible` and may be `qtox-busy` too.
What about the `qtox-event*` icons? They are different from qtox originals.

wa4557:
You could make `qtox-offline` 40% transparent, and colour `qtox-event` Numix warning orange. Apart from that I think they are quite nice :)

frisbee00:
Numix warning orange? As in https://github.com/numixproject/numix-icon-theme/blob/master/Numix/24/status/battery-010.svg?
I had assumed monochromatic to be dark or light. Since other single colours are possible, will change the `qtox-event` icon to match the original - same as `qtox-online` but coloured.
Anything for `qtox-event-small`?
Will make the `qtox-offline` 40% transparent.

wa4557:
There are 3 possible colours for Numix tray icons:

- #ececec (#353535 for dark variant); main colour
- #ef555c for errors
- #ff8000 for warnings /reminders
- #54cdee for notifications

Your icons should consist of these 4 colours, no more

frisbee00:
So how about this - to be consistent with the original qtox icons, use `qtox online` for both `qtox-event-small`and `qtox-event` but coloured in #54cdee and #ff8000 respectively.

wa4557:
Yeah, that's good

frisbee00:
- Changed stroke paint from None to Unset

wa4557:
Now the annoying part: 

- Move the icons from 24/panel to 24/status (panel icons are a special case)
-  you have to add the icons also  to 22/status and 16/status
    - 22/status is easy: just copy the same icon as for 24/status to a 22px canvas
    - 16/status: Maybe the icons also fit on a 16px canvas otherwise you'll have to align them to grid on a 16px canvas as well
- all the icons (except for the coloured ones) have to go into Numix-Light as well. Just copy them there and replace #ececec with #353535

frisbee00:
- Moved icons to `status`
- Added 22px and 16px icons
 - You might want to check the 16px icons; they are not perfectly scaled down versions of 24px - tried to align to the grid
- Added icons for Numix-Light

frisbee00:
- Removed `qtox-event-small` since they are unused as per https://github.com/qTox/qTox/pull/2689
- Changed colour of `qtox-event` to #54cdee (notification)

wa4557:
The 16px icons could be a tad bigger (no need to directly resize them down from 22px), but they are hardly ever used anyways. You should also remove `qtox-event` from Numix-Light, since they are inherited from Numix. 

Apart from that they look pretty fine :)

frisbee00:
Grid alignment would be worse if the height of the path is increased to 14px (assuming next even number would be preferred). Proceed?

wa4557:
Nah I guess 14px is fine :)

frisbee00:
OK. Will work on it.

frisbee00:
- Increased height of paths in 16px icons to 14px from 12px

Foggalong:
@frisbee00 These designs are looking good, but there are still a few cases where the objects inside the lock circle cutout aren't aligned properly to grid. It may looks a bit weird to malform the shape when doing so in Inkscape but the result will look a lot better when viewed at the actual image size.

frisbee00:
The circle in 22px and 24px icons are 7px in diameter and the shapes inside are not fully aligned to grid. I guess those are the ones you are talking about. I thought aligning them more to the grid would need an even number of pixels as the circle diameter. I had tried 6px (felt too small) and 8px (too big).

Any ideas?

Foggalong:
There's also the no-entry one at 16px, but there might be others I've missed. You can get away with the 7px circle by fudging the shapes a little bit. Even if it's not a perfect semi-circle it'll both look better and still like a semi-circle when viewed at 16/22/24px

frisbee00:
No-entry one?
The one with semi-circle would be `qtox-away` icon. Will this be fine?
test

Foggalong:
Just looking at that screenshot, I've realised the circle elements themselves should also probably be aligned to grid. It doesn't make as big a difference as straight edges but it still makes for an overall neater icon.

frisbee00:
OK. Let's try the 8px circle with modifications to the callout symbol.
test

Foggalong:
This would also apply to the circle making up the speech bubble so it might be preferable to have the slightly smaller circle down there.

frisbee00:
Like this?
test

Foggalong:
@frisbee00 exactly like this!

frisbee00:
And the 16px icons...
test

Foggalong:
@frisbee00 Almost, but the sides of upper part of the lock needs to either be wider or narrower so they align to grid and the circle part of the speech bubble could be made 4px diameter for neatness.

frisbee00:
This?
test

Foggalong:
Bottom one :+1:

frisbee00:
Pushed the modified icons.
What about the svg cleanup part? I found a script here:
https://github.com/numixproject/numix-tools/blob/master/numix-tools/svgclean
Should I use it with the `-f` option?

wa4557:
Installing svgcleaner is hard on some distros. But yeah, if you manage to install it that would be great :)

frisbee00:
No problems here. `-f` or not?

wa4557:
No

frisbee00:
Not all of them were cleaned up using the `svgclean` script. Is that how it is or did I miss something?

wa4557:
Try with the `-f` argument then

frisbee00:
All of them seem to be cleaned up with `-f` argument.

wa4557:
@frisbee00 I on e told you to remove coloured versions of your Icons from numix-light. Can you symlink them Back in? (Using relative path symlinks to the Sam Icons in numix)

Foggalong:
Job well done :+1:

frisbee00:
Sorry for commenting on the closed request, but the following things came up.
- The tray icons are visible when setting the icon theme to Numix but not when using Numix-Circle (totally missed this while testing). It works when the icons are moved to the panel directory. Inheritance issues?

- My fellow users here have complained about the following things **at small sizes**:
 - The circle forming the status symbol looked merged with the callout symbol
 - The corners of the lock are too sharp

I tried reducing the circle diameter to 6px from 8px and increasing the corner fillets to 2px.
Current: 
current
Modified:
alternate

Any comments?

wa4557:
Yeah the panel directory was made exactly for that, you should/can move them (but probably wait until #1134  is sorted out before opening a new PR)

I prefer your modified version

PR #1123, opened 2016-09-02
A few symbolic icons from Adwaita

wa4557:
A few from #577


Foggalong:
@wa4557 According to the original issue, in `media-playlist-consecutive-symbolic-rtl` the `-rtl` should come before the `-symbolic` not after.

wa4557:
I have actually both in my current Adwaita installation ...

Foggalong:
The one you've just commited is a different order again - is that order also present in your installation? For clarification `media-playlist-consecutive-rtl-symbolic` was from the original issue, `media-playlist-consecutive-symbolic-rtl` is one you added in this pulls original commit, and `media-playlist-rtl-consecutive` (with no `-symbolic`) is the one added in the second commit of this pull.

wa4557:
Oh sorry, the `-symbolic` should be there of course. 
So we need `media-playlist-consecutive-rtl-symbolic` and `media-playlist-rtl-consecutive`

Foggalong:
Cool, so it's just `media-playlist-consecutive-rtl-symbolic` that's missing now?

wa4557:
There are definitely too many hyphens there. The one we need are `consecutive-rtl-symbolic` and `consecutive-symbolic-rtl`

Foggalong:
Oh right! Sorry, I misunderstood. Thanks :+1:

PR #1124, opened 2016-09-03
A few symbolic icons from Adwaita

wa4557:
From #577.

EDIT: I also added a few non symbolic icons


Foggalong:
@wa4557 A few of these icons aren't showing up in the GitHub render which iirc means something's wrong with the icon? They're also behaving a bit weird when opened in Inkscapes, the shapes appearing bigger than they actually why. I've had a quick play around but can't work out why :/
shot1

wa4557:
Hmm. Cannot reproduce, for me every icon shows correctly up in github. What icon are you talking about?

Foggalong:
Both `media-zip-symbolic` and `uninterruptible-power-supply-symbolic` are still showing up blank so I'm guessing there's a mask there too?

wa4557:
Alright. Now it should be fixed. Did show up correctly in chrome, though, I guess you are using firefox :)

Foggalong:
Aye, Mozilla all the way! Was it a mask problem or something else causing it?

wa4557:
some leftover gradient definitions, should  all be gone now.

Btw: since all PR are closed now we could fix the ppa

PR #1125, opened 2016-09-05
Fixes #989

### LINKED ISSUE #989: nm-applet cellular icon is broken

wa4557:
I used a font that enables to read the letters even at smaller sizes. Looks decent at original size


Foggalong:
I'm a little apprehensive about using a different font style to what we use elsewhere, but willing to give it a go to fix such an ugly bug.

wa4557:
Me too, it's just completely unreadable if we use the same font for such a small size. As soon as this changes we should switch back to the other icons...

Foggalong:
Agreed. Your original icons were sweet :heart:

PR #1130, opened 2016-09-06
Two icons for elementary

wa4557:
From #437

P.S.: There are more icons from my last few PRs that can be ticked off that list.


Foggalong:
Thanks, think I got them now :+1:

Issue #1025, opened 2016-03-22
XFCE panel nm-applet VPN connecting and connected icons are broken.


James5979:
I think this may be a problem with overlays, similar to: https://github.com/numixproject/numix-icon-theme/issues/989 .  Took some screenshots to help:

Normal USB tethered connection, same as normal Ethernet connection for icons:
https://i.imgur.com/Q0vXJqU.png

VPN connecting:
https://i.imgur.com/I8cREj9.png

VPN connected:
https://i.imgur.com/EXc0z9F.png

The last two screenshots above show the broken looking nm-applet icons. Thanks for anything that you can do to remedy this.

wa4557:
Well that's new. For the second screenshot is there an animation showing?

James5979:
I made a .gif file of the animation of the icon when connecting to the VPN.

https://i.imgur.com/4HoDSxw.gifv

wa4557:
I'm not sure we can do anything about this :( In order for this to work out, we would need to change the animations. Which would be very easy to do, however, the thing is Ubuntu for example is not using those overlays, so removing the wifi shape from the icons would show just emptiness. 

I'm not a big fan of those overlay icons, they just make everthing more complicated...

James5979:
Yes, that would not be acceptable for Ubuntu users, so this leaves us in a bit of a pickle. As a suggestion, a possible way to overcome this would be to make a clone of the numix icon theme specifically for Xfce, but that would be overly complicating for just a single bug. The only possible way (maybe), would be to make an Xfce specific fixer script that an Xfce user could run to replace the broken animation with the correct annimation, but still not an elegant or concise solution. Still, thankyou for trying to resolve this issue and I hope one day Xfce would stop using overlays in their nm-applet. Keep up the good work, appreciate all that you guys do.

jrb:
Unfortunately this issue isn't limited to just Xfce, nor is it only the VPN. When you throw additional Mobile Broadband into the mix it gets even messier as Numix doesn't yet have an LTE icon:

![numixlte](https://cloud.githubusercontent.com/assets/397735/18452158/aebbd3f4-78ee-11e6-927f-6c8731d9ac6d.png)

![numixltevpn](https://cloud.githubusercontent.com/assets/397735/18452163/b4564772-78ee-11e6-8a46-3528e3d82880.png)

After digging through the network-manger source a bit, it looks like the possible icons are:

```
Disconnected:
    nm-no-connection

[...]

PR #1135, opened 2016-09-12
nm-applet compatible vpn connecting animation.

jrb:
I put together a new set of animations for nm-vpn-connecting, using the existing nm-vpn-lock with varying opacity. This appears to resolve the vpn related aspects of #1025. 

I'm also putting together a version of nm-tech-* which moves the access tech indicator to the left of the signal indicator, and sets nm-wwan-tower to completely transparent. I'll open a PR for those if this is well received.


wa4557:
Wait there are nm-tech Icons in numix already. Maybe we need to move them to the left side though

wa4557:
Your PR is alos problematic, because of the same issues raised in #1025. On ubuntu this leads to a situation like this:
unbenannt

we have to decide what we want to fix. XFCE and Unity/GNOME is obviously not possible

jrb:
Yes, sorry, that's all I meant. I was just going to flip all the current tech icons from bottom right to bottom left, so they would avoid the VPN icons. It looks like the LTE access icon is currently missing, so that may need to be added.

wa4557:
Isn't there also a 4G icon (i.e. LTE?)

jrb:
Ah, that behavior in Unity/Gnome is unfortunate. For what it's worth, I'm in Mate, not Xfce, and see the Wifi icon with VPN connecting overlaid. Maybe the way to go is an overlay theme that depends on Numix and overrides parts of nm-*

There is a 4G icon, but LTE is an independent tech in NM. I suppose a symlink is sufficient. I want to say 4G gets used for WiMax connections while LTE gets used for LTE.

wa4557:
But an overlay of WiFi icon with von Lock is fine, isn't it? The Problem in #1025 is that the wired icon is overlayed

jrb:
I think the base image is the wired icon, with the vpn connecting animation layered on top. Because vpn connecting animation is white, and makes use of transparency, the solid 'wired' icon appears to be on top.

After looking more closely at the image in #1025, you're right. This PR would only partially resolve the issue, as the positioning of the lock itself conflicts with the wired icon.

I'm not sure how this would be resolved short of either moving the lock or altering the wired image. :-/

jrb:
A bit more digging through the network-manager-applet source makes it look like this "Gnome/Unity vs XFCE/Mate problem is related to use of libindicate on the Gnome/Unity side, and an old school status icon for XFCE/Mate. 

This PR can probably be closed. :-/ Thanks for your help!

wa4557:
Well let's hope xfce and mate change to libindicate as well at some point...

PR #1145, opened 2016-10-03
Makes nm-wwan-tower-transparent

### LINKED ISSUE #989: nm-applet cellular icon is broken

wa4557:
This should now really fix #989. There are actually 4 icons overlayed, this is extremely stupid...


Foggalong:
Just when you think you've seen it all with stupid icon setups, something new comes along and blows it out of the water.

PR #1157, opened 2016-10-30
sync numix with breeze icons

DarkknightAK:
added the symlinks which are available in breeze but not in numix. first simply sync


wa4557:
That's actually really nice :) I haven't looked through all yet, but for most icons that should be fine. One thing though, different to breeze Numix has action icons for all sizes up to `48px` you should add those symlinks there too.

Btw.: If you want to contribute icons for breeze later on (not symlinks), I can support you in making icons for all sizes, if you provide the initial design...

DarkknightAK:
Good to know thanks

palob:
Is this ready to go?

PR #1185, opened 2017-02-19
Added symlinks for xfce4-power-manager

siema:


Foggalong:
Just to check in with you @dirtydancing because this is your area of expertise: will this mess anything up?

bil-elmoussaoui:
@Foggalong Can you include this on the new release? we will get faster review this way :)

PR #1204, opened 2017-04-23
Fixes typo for syncthing

wa4557:
I was wondering why the icons don't show up. Turns out we had the icon names wrong the whole time :)

See here:
https://github.com/syncthing/syncthing-gtk/tree/master/icons/24x24/status


TiZ-EX1:
There is one symlink still using the old name: si-idle.svg -> si-syncing-0.svg. This causes the custom idle icon not to appear.

TiZ-EX1:
Thanks guys!

PR #1206, opened 2017-04-25
Panel icons for Xfce4 Mailwatch Plugin

frisbee00:
Added two panel icons for Xfce4 Mailwatch Plugin
- `xfce-newmail`
- `xfce-nomail`


frisbee00:
Can you please hold this merge. I think there is problem with location of icons; will confirm and get back.

frisbee00:
The locations are fine. Also removed the scaling factor accidentally set while opening the original icons from Xfce4 Mailwatch Plugin in Inkscape.

PR #1210, opened 2017-04-30
A few inkscape icons

wa4557:
Two more icons to complete inkscape icon theme (stil a lot to do). Also from #580


wa4557:
Starting to look good :)
unbenannt

PR #1219, opened 2017-05-20
More inkscape icons

wa4557:
Even more inkscape icons.


wa4557:
@Foggalong @palob @bil-elmoussaoui 

What do you guys thinking about redesigning those icons?
  
transform-scale
transform-scale
transform-shear-*
transform-shear-down

to a design similiar to this one (i.e. white rectangle with grey stroke and only the arrow in purplish color)?

transform-move-vertical
transform-move-vertical

This would make the transform-* icons consistent (if you like the new proposed design, oc)

wa4557:
For reference these are the proposed (re-)designs:
transform-move
transform-move-horizontal
transform-move-vertical
transform-scale
transform-scale-horizontal
transform-scale-vertical
transform-shear-down
transform-shear-left
transform-shear-right
transform-shear-up

palob:
Grey strokes might be problematic on a greyish GTK+ theme (there's a lot of different shades out there). I think I pefer solid objects to contoured ones because it makes for a less fine-grained/less busy look.

Two different colours provide a good distinction between objects and actions but do you think keeping it to one colour (different shades of blue or whichever colour) makes any sense?

wa4557:
I see your point. Even if the gtk theme colours matches the countour colour this would still work, though since we have the white filling.

Don't you think that this looks much better 

screenshot from 2017-05-22 11-14-27

than this?
screenshot from 2017-05-22 11-15-11

wa4557:
I've added three more icons (less controversial ones :) Still want to make all the `transform-` icons consistent (at least colorwise). If you guys prefer I can fiddle around to make icons with no stroke, though we have actually a lot icons in `actions` with strokes...

palob:
Well, you do all the work here, so I'll trust you find a way :)

By the way, do you know why the icons in the lower screenshot look blurry?

wa4557:
Ok. So no stroke it is then?

Yes there is a symlink missing for the 24px size, therefore the 22px icons were used. I will add them once I add the corresponding icons.

palob:
It has not beend decided yet I think.

Did you take the icons for a ride together with some different GTK themes?

wa4557:
Will do

wa4557:
Any change requests?
screenshot from 2017-06-04 13-15-17

bil-elmoussaoui:
They look awesome!

wa4557:
I have more icons, but I guess it makes sense to wait and open another PR if these icons are good to go

palob:
Feel free to open another PR or show some screenshots.

wa4557:
Alright, here's what I got so far. 

The ruler needs a redo, it doesn't look good.
screenshot from 2017-06-11 12-35-37

palob:
Wow, great times ahead!

wa4557:
Unfortunately inkscape uses a weird 18px*18px icon size in the top panel. This makes every icon there blurry; arguably we don't want to open the can of worms and add 18px icons. Not sure how to deal with that right now...

wa4557:
Just for anyone that likes crisp icons in inkscape:

Creating a file in `/etc/gtk-2.0/gtkrc`

with content 

```
gtk-icon-sizes = "inkscape-decoration=16,16"
gtk-icon-sizes = "gtk-small-toolbar=16,16"
```
solved the issue for me, such that inkscape uses 16px icons or larger everywhere. Trying to figure out if this can be included in the gtk-theme

palob:
Oh no. With the Inkscape GTK3 port underway at long last and assuming upstream won't get new icons how is this to be dealt with in future?.
Maybe we do need 18px icons (16px ones with empty canvas around).

wa4557:
See my last comment. This can be fixed manually or in the gtk theme. With gtk3 this should be no problem anywas

palob:
Most users won't fiddle about with the gtk theme and many of our icon users don't use Numix GTK.

Why don't you think this will be a problem with GTK3?

wa4557:
Looking at the way they want to implement GTK3 support (inksape wiki) they are going to outsource  all the icon stuff to gtk, instead of doing all manually. So the icon lookup should be good

palob:
So the icon size isn't hardcoded?

wa4557:
Right now? No. Just the default value is weird, but can be overwritten in the gtk-theme. 

In the future with GTK3 the problem should be gone, because the default values are then correct because given by GTK directly

palob:
@Foggalong How to go on? Solid colour icons or contoured ones?

Foggalong:
@wa4557 @palob I'm worried about thing getting ugly when the contoured colour matches that (even closely) of the GTK theme, but the design is a big improvement. Is there anyway to compromise between, perhaps using a colour other than grey? I'm not sure how that'd end up looking

wa4557:
I tried a few different themes and I don't see a big problem with the grey (note that also Breeze is using grey for all there action icons)

Numix:

numix

Arc:
arc

Arc-dark
arc-dark

Ambiance:
ambiance

Greybird:
greybird

wa4557:
@Foggalong So what's the verdict? Is this good to go from your point of view, or do you think I should change something?

PR #1239, opened 2017-08-25
Fixes #1237

### LINKED ISSUE #1237: Questionable design for indicator-stickynotes-mono
Looks like this.

stickynotes

I suppose it is intentional but I think we should streamline it.

palob:
Fixes #1237


bil-elmoussaoui:
It seems that you forgot the 24px icon of Numix Light :P

palob:
Good catch :-D

Issue #1238, opened 2017-08-18
Issues with RSS related-icons


palob:
The symbolic icon (is it ``application-rss+xml-symbolic``?) looks like this in real life (screenshot from Vocal).
![rssicon](https://user-images.githubusercontent.com/7050624/29447892-516bfe88-83f4-11e7-9f46-ab392a3db2c2.png)

Probably because of strokes rendering issues.
![strokes](https://user-images.githubusercontent.com/7050624/29447959-a1321344-83f4-11e7-961c-026169bc38ad.png)

For some sizes the mimetype icon ``application-rss`` needs realignment.

All RSS icons should be redesigned in line with the symbol on the app icon.

wa4557:
@palob The symbol is the same as in the app icon though, isn't it? (`rss.svg` in numix-circle)

palob:
Yep, only other sizes.

PR #1240, opened 2017-08-25
Fixes symbolic RSS mimetype

palob:
For #1238


palob:
I'll do the mimetypes later. There not as bad as I thought and not as easy to fix as I thought.

PR #1245, opened 2017-09-17
Adds synergy status icons

wa4557:
Tray icons for synergy, a mouse sharing app. Unfortunately hardcoded but Hardcoded-Tray fixes them


wa4557:
No. It just is the standard icon when everything is fine but nothing connected. The other one (with `ok`) when a client is connected

Foggalong:
That's such a bizarre icon. Is that the same iconography synergy uses?

wa4557:
I know. That's the original icon:
https://camo.githubusercontent.com/718870c23b7e56cc34bf493ad626f0584ee8a756/687474703a2f2f736c31706b6e30372e7774662f73796e657267795f72322e737667

Didn't know what to make out of it :/ (I'm also a bit rusty in making icons...)

Foggalong:
How weird. Well if that's their icon that's their icon, maybe it has some link to the purpose that isn't evident at face value. It's not a _bad_ icon on their end, just unexpected

wa4557:
Cool :+1:

PR #1249, opened 2017-10-29
Provide message-indicator-symbolic

### LINKED ISSUE #1248: Gnome-Shell notification indicator icons looks ugly
It used to be that the Gnome-Shell notification indicator was a monochrome solid white circle, indicating a notification. However, something changed recently and now the notification indicator is a full color icon that doesn't fit at all. Going back to Adwaita icon theme, the solid white circle comes back, so it makes me think that for a while numix was defaulting to the adwaita icon or something.

Is there a way to go back to default to adwaita, or perhaps coming up with a better, more monochrome, notification icon?

palob:
Fixes #1248

Just a 4x4 circle like it looked before. Will decide what to do with ``message.svg`` later.


wa4557:
Numix-light icon is missing

palob:
Thanks, added it.

PR #1251, opened 2017-10-30
Fixes aligning and canvas size for bluetooth icons

### LINKED ISSUE #1235: Bluetooth icon is misaligned and missized
Hello!
I noticed that whenever I have blueman-applet on, the system tray icon is a bit off shown here:
screenshot_2017-08-15_08-14-31
screenshot_2017-08-15_08-14-00

I was wondering if there is any way that I can get them fixed?
(BTW I have already posted an issue for blueman and they said to come here!)

wa4557:
Fixes #1235


palob:
Redesign is fine but I think we should get rid of the lock metaphor. Apart from being potentially confusing it looks busy.

wa4557:
What about the scalable icons? Should I go there with the "new" `16px` design?

PR #1252, opened 2017-10-30
Cleans up RSS icon

### LINKED ISSUE #1238: Issues with RSS related-icons

wa4557:
Fixes #1238


palob:
Thanks :+1:  how does it look with the symbol up 1px (maybe more for bigger sizes) for 24px onward?

wa4557:
I think you meant 1px down? I think it's worse:

current:
  

1px down:
  

1px up:

palob:
Yeah, down of course. I think it doesn't clash as much with the dog ear this way.

wa4557:
Alright, going with one px down then

palob:
I think all mimetype symbols are slightly off-centre.

wa4557:
You're right, that might be a bit harder to fix...

wa4557:
Only the 22px / 24px variant is off-centre. But to keep it aligned to grid we have to keep it that way...

palob:
No, I was talking about the existing ones :) Most of the symbols are shifted down due to the dog ear.

PR #1265, opened 2017-11-11
Changes design of user-trash-full

### LINKED ISSUE #1014: Trash icon is not changing when trash is not empty
In previous versions of Numix Circle Theme trash icon were colored to red, when trash wasn't empty. Right now, it's gray all the time even if something is in the trash. I think that different color (e.g. red) or different icon for non-empty trash is useful for me as a system user. Can you bring previous behavior back or add similar behavior (different color or icon for non-empty trash)?

screen-01
screen-02

wa4557:
SHould be discernible enough. Definitely better than the current situtation. Fixes #1014


palob:
I wonder however if the lines on the full one should be straight as well. Right now they almost look straight with a little bit "muck" around.

Foggalong:
Why not keep the red colour on the full icon for those DEs that still support it?

wa4557:
I changed my last commit. Straight lines and keeping the color

PR #1273, opened 2017-11-26
Adds panel icons for xfce4-dict

frisbee00:
Just check the icons. Any better ideas?


frisbee00:
OK. I have changed the icons to look similar to app icons for `goldendict` or `stardict` from Numix. The other option is to make it looking similar to `accessories-dictionary` which is the app icon for xfce4-dict with Numix-Circle (which is basically **Aa** in a rounded square).

PR #1281, opened 2017-12-25
Adds new teamviewer icons

wa4557:
Teamviewer is now a native linux app (yay), and uses only one panel icon for now. Added them

### DIFF
--- a/Numix-Light/16/status/TeamViewer.svg
+++ b/Numix-Light/16/status/TeamViewer.svg
@@ -1,0 +1,1 @@
+teamviewer-indicator.svg
--- a/Numix-Light/22/status/TeamViewer.svg
+++ b/Numix-Light/22/status/TeamViewer.svg
@@ -1,0 +1,1 @@
+teamviewer-indicator.svg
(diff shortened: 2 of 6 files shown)


Foggalong:
If Linux file names weren't case sensitive we wouldn't be able to merge this because the app icon name is `teamviewer.svg` :sweat_smile:

bil-elmoussaoui:
Are you sure about this?  In my case, using teamviewer beta, the icon still doesn't look correctly. 
screenshot from 2017-12-25 13-43-14
screenshot from 2017-12-25 13-42-57

wa4557:
Yes you're right. I guess we have to move them to the `panel` folder

bil-elmoussaoui:
@wa4557 I don't think it the icon is not hardcoded. all the icons are links to /opt/teamviewer/tv_bin/desktop. If you replace those icons there, nothing changes at all :( 
+ The TeamViewer is the new icon name for the application itself.

wa4557:
In my case, when put the icon in the panel folder everythings fine. The app icon is shown in the launcher, the monochrome icon in the tray... Maybe it's different in GNOME?

PR #1305, opened 2018-05-08
Some symlinks for KDE 5 / dolphin VCS emblems

jlu5:
These emblems are used by [Dolphin's VCS and Dropbox integration plugins](https://cgit.kde.org/dolphin-plugins.git/tree/). Without the applicable icons or icons to take their place, all of these emblems show something like the [unknown file type icon](https://github.com/numixproject/numix-icon-theme/blob/master/Numix/22/mimetypes/unknown.svg) instead (which in itself is a weird fallback IMHO...)

These links are not one to one matches (e.g. the "locally modified" and "update required" icons), but still preserve the original emblems' intents to some degree. I am not a designer myself, so I don't think I can do much more here.

A quick comparison with Breeze:
numix-vcs-symlinks


Foggalong:
This was great! There was one small issue which wasn't your fault, just some error from the previous name structure of the remove emblem icons. I fixed that in that last commit so it should all be good now :+1:

PR #1317, opened 2018-11-05
Make view-sidebar-symbolic unique

johnbeard:
Previously this was the same as open-menu-symbolic,
which was confusing in places like evince's toolbar,
where both icons existed.

2018-11-05_131045_321x90_screenshot


bilelmoussaoui:
You need to update the Numix Light version too.

johnbeard:
Light version updated.

palob:
It appears Inkscape/scour have left some cruft here. Can you manually remove the ``;stroke-width:.77461`` inside the last ``style="...."`` declaration?

johnbeard:
Done.

PR #1321, opened 2019-01-05
Add panel icons for enpass6

bernimoses:
Will this be merged into Numix-Circle, ...?


palob:
No, into the Numix base theme. But that's the right place :)

PR #1328, opened 2019-05-24
mimetype icons for Groovy source files

palob:
groovy


palob:
@Foggalong This is the missed one :)

PR #1348, opened 2020-03-03
Mimetype icon for Erlang source files

palob:
text-x-erlang


Foggalong:
I thought this was a graphical glitch until I looked at the upstream icon :joy:

PR #1353, opened 2020-04-03
Relink troff mimetype icons

palob:
(and add missing symlinks).
Previously this was linked to the ebook mimetype icon which doesn't make sense.


Foggalong:
I'd never heard of Troff before, that sounds cool if not very esoteric. For anyone else interested [this page](https://troff.org/faq.html) made for informative reading.

PR #1356, opened 2020-04-06
Removes unncessary padding from 16px icons

### LINKED ISSUE #1355: Ubuntu 20.04 16px status icon too much padding
Ubuntu 20.04 will use 16px icons instead of 22px icons in the top bar. Since all (most?) 16px status icons have way too much padding that looks as if these icons were way too small. I think a proper solution would be to go through the 16px icons and remove the padding

See here for reference: https://github.com/ubuntu/gnome-shell-extension-appindicator/issues/211

I have some free time the next few days, I could go through them and change it?!

andia89:
That should fix #1355 . A lot of icons were not aligned to grid (I tried to fix that where possible) and I also found one icon who should be symlinked instead of duplicated. apart from that I tried to remove as much padding from 16px icons as possible


andia89:
I did. But it wasn't that many

Foggalong:
```
> andia: "It wasn't that many"
> also andia: submits a PR changing 515 files at once
```

Foggalong:
But seriously, great job with this! Very thorough PR and nice catch with the Nvidia duplicate too :+1:

PR #1360, opened 2020-05-07
Mimetype icons for Arch Linux Packages

palob:
application-x-alpm-package


Foggalong:
I'd not realised there was actually a mime association with Arch's packages, I'd always just assumed they were straight tars. It's also quite difficult to find any information about them since searches with "ALPS", "Arch", and "Arch Linux Packages" often bring up ALPS the Linux touchpad software or [alps](https://aur.archlinux.org/packages/alps/) the quantum mechanics simulator.

PR #1373, opened 2020-08-02
PCB file mimetype icons

palob:


Foggalong:
Very cool design, love the colour choice!

PR #1376, opened 2020-08-02
mimetype icons for excellon drill files

palob:


Foggalong (approved):
Nice, I really like this one!

PR #1400, opened 2021-08-15
Add GitHub folder

drakkar1969:
Opening a PR here to include Github folder.

Reflects the PR previously opened and merged in the numix-folders project: https://github.com/numixproject/numix-folders/pull/195.


palob:
Icons sizes 22 and 24 have the symbols misplaced vertically ever so slightly.

drakkar1969:
Vertical positions should be fixed now

palob (approved):
Thank you, positioning is fixed. There are still some misalignments left inside the symbol but right now the icons look exactly like the ones in *numix-folders* and fixing those will probably distort the symbols in some cases so it's alright.

For future reference: For new additions it is preferred to have the objects composing an icon ungrouped as this saves a little bit of disk space/rendering overhead.

PR #1401, opened 2021-08-16
Add torrent folder

drakkar1969:
Add torrent folder for all sizes/colors

folder-torrent

Uses `*-folder-system` as reference for placement of torrent resource


palob:
Hi, thanks for this huge contribution.

+ Sizes 16x16, 22x22, 24x24, 48x48 need proper alignment of the symbols to the grid. Right now it looks like this.
misalign16
The symbols will hve to be redrawn from scratch using a 1px/2px/3px.... wide stroke.
Sizes 32x32 and 64x64 are fine.

+ All symbolic links ``folder-torrent.svg``are broken

drakkar1969:
Sorry for the huge delay
Second try complete, 16x16, 22x22, 24x24, 32x32 and 48x48 sizes redone to fix alignment

palob (changes requested):
Thank you for doing this tiresome work.
Alignment is fine now. There is a problem however.

You used clipping for the symbols which we normally do not accept because it doesn't work in Qt-based DEs (clipping isn't supported by QtSVG).
You can check this by viewing the files in a Qt-based image viewer like Gwenview or Nomacs.

Will it be a lot of extra work to convert the clippings by means of path intersection?
Right now the icons do not look broken but different, not strictly like the torrent symbol.

drakkar1969:
Thanks for the feedback, I had no idea about the issue with clipping with Qt. Bit of an inkscape novice here :).
I have fixed (I think) the issue for the 64x64 size. If I am on the right track, happy to fix the other sizes too

drakkar1969:
All sizes should now be ok

palob (approved):
Thank you, that's exactly what was needed. No traces left, as far as I can see.

PR #1410, opened 2022-02-23
Add elements tray icon

### LINKED ISSUE #1408: Element tray-icon (formerly riot)
element is currently not using a numix-tray-icon, because there is none.

I highly believe, that it is supported by [`Hardcode-Tray`](https://github.com/bilelmoussaoui/Hardcode-Tray/blob/master/data/database/element.electron.json).
An inspiration can be seen at [papirus-icon-theme](https://github.com/PapirusDevelopmentTeam/papirus-icon-theme/blob/master/Papirus/24x24/panel/element-desktop-tray.svg). is it possible to use the same files? or how are the files generally created?

only two file need to be created:
```
Numix/24/panel/element-desktop-tray.svg
Numix-Light/24/panel/element-desktop-tray.svg
```

Stunkymonkey:
Fixes #1408.
I have never done an SVG before. I downloaded: https://en.m.wikipedia.org/wiki/File:Element_(software)_logo.svg and edited the details by hand in the file (adjusted colors, removed useless information). Hope everything is as expected.
ping @palob


palob (changes requested):
Thanks for entering the SVG business :-)

At this 24px size the symbol is slightly misaligned w.r.t. the pixel grid (`#`key toggles it in Inkscape).

misaign

To redraw a aligned symbol in Inkscape you can draw a circle using the circle tool (16px X 16px, placed at X=5,5px ; Y=1,5px) and add a 3px wide stroke (just set a stroke colour in the *Fill and Stroke* toolbox. Stroke width can be set in the *Stroke Style* tab. Stroke cap should be round).

circle

(added transparency for illustration purposes)

Now select the circle and make it extend from `90 deg` to `180 deg` only and switch to *arc*.
circle-settings

Remove the fill colour and convert the result into a path. `Path -> Stroke to Path` in the menu.
Now just duplicate 3 times and rotate+move to the correct positions.

Feel free to ask back. If you don't feel like doing this I can take over.


circle

Stunkymonkey:
I just copy/pasted the original icon. The things you mentioned with inkscape are out of my experience.
I tried the setps you provided, and understand what is going on.
Sadly the placement (X,Y) in the beginning is not looking correct for me.
I expect it to be my fault. So please continue the work. You will solve this much faster.

palob:
By the way... 24px is this correct? No 16px icons?

Stunkymonkey:
maybe we have to add both? I am not the expert here. I just saw the `redshift-status-on.svg` in the `24px` folder i expected it. But it also exists in the `16px` folder. No idea what is correct :man_shrugging:

palob:
I don't use any applications which show tray icons in GNOME but it doesn't look like the tray could accomodate 24px icons.

Stunkymonkey:
when using `sway` as window manager the bar can use 24px height.

Stunkymonkey:
the problem was, that inkscape always remembers the settings from before. I did not know how to start properly.
Currently I have an image like this: the black is the ones i was doing, and the green the old one.
element

if you want to have a closer look: 
element

is everything right? i would continue to adjust the colors and make it with different sizes

Stunkymonkey:
to be honest I do not see a problem in using the original file in this PR

Stunkymonkey:
or do you mean the `viewBox` is the problem? and we should better rescale it?

palob:
I'll have some time to fiish this up tomorrow. I can add 16px ones as well. Looking at our current panel icons they all don't use the full 24x24 canvas and have a 2px padding on all sides. Our 16px icons have no padding.

Other icon themes (e.g. Papirus) also use at least 2px padding for the 24px panel icons. Do you currently use your custom icons? Do they look bigger compared to other panel icons? How does the original non-themed icon look?

I'll post an explanation for that alignment thing soon.

palob:
Here you see 2 rectangles. The upper one is aligned to the pixel-grid ("pixel-perfect"), the other one isn't. Take a look at the upper red-white border of the lower rectangle

align1
 There is a line of pixels with two different colours inside. However the physical pixels of a display can only show one single colour each. This colour will be calculated by weighted averaging.

On a display the misaligned rectangle will look like this

align2

leading to a blurry impression. 

The following picture shows two circles. The top one will look crisper on a display.

align3

Stunkymonkey:
> blurry impression

thats what I was missing. I did not think of that. But thanks for taking over

Stunkymonkey:
> Do you currently use your custom icons?

I currently do not use my icons, since they are hardcoded.

palob:
Ah, so it's a case for Hardcode-tray?

palob:
Added a 2px padding for the `24` size and new `16`size icons. Boy, there are really bad icons w.r.t grid alignment and other technical/design aspects among our existing panel icons.

PR #1413, opened 2022-04-11
Add the conda-forge installation guide in the README

GiulioRomualdi:
Hi all maintainers! First of all, I would like to thank you all for the effort in maintaining this very complete set of icons!

I packaged `numix-icon-theme` in conda-forge: https://github.com/conda-forge/numix-icon-theme-feedstock

This PR updates the readme with the installation command that should be used to install `numix-icon-theme` with conda. Thanks to this the package is available not only on Linux but also on macOS and windows.


palob:
Thank you for maintaining the package. There is one small error in the description in the package readme. I think the license is ``GPL-3.0-or-later``.
I see you also fixed the dead distro icon links. Where exactly are those sourced from? Will the links be more robust this way?
Fedora icon shows their old logo.

PR #1425, opened 2022-10-11
Adding icons for subscript and superscript.

geggi:
A proposal for two missing icons, i.e. subscript and superscript.


palob (changes requested):
Thanks for your contribution. 

Can you please tell me where those icons are used? Judging from the names those aren't symbolic icons which means the icons should be present in different sizes and be coloured/multicolour icons (we use blue for our other ``format-**`` action icons).

+ The subscript/superscript **2** must be converted to a path, otherwise it will look differently depending on whether the user have installed the font you used (and depending on other SVG rendering internals).

+ Can you please merge the two paths the **a** is comprised of and remove unnneeded nodes? I know this bug is already present in existing icons.

+ Please save the icons as "Optimised SVG" from Inkscape with options as laid out here:
https://github.com/numixproject/numix-core/wiki/Optimise-Options

+ Currently the commit message is the generic ``Add files via upload``. Do you know how to write/edit git commit messages in order to have something meaningful?

Don't hesitate to ask if you have further questions.

geggi:
Thanks a lot for the good and detailed feedback.

I want the icons to be used in Gramps project: https://github.com/gramps-project/gramps/pull/1266
The new icons are based on gnome legacy stencils (https://github.com/gramps-project/gramps/files/6841798/gnome-stencils-legacy.super-sub.zip)

I tried to follow most of your comments (new version committed).
Few questions:
- I couldn't open the svg files in the 16, 22, etc. folders.
- why are different sizes needed? i assume as it is a vector graphic it can easily be scaled.
- I have provided a more meaningful commit message. Please advise if something different is needed (I am not very familiar with git/github)

geggi:
Thanks a lot for the detailed explanation. That makes absolute sense.
I will check my icons regarding the mentioned pixel perfect design.

> Do you mean in Inkscape or an image viewer?

I had tried to open the file in Inkscape.

PS: I am not a graphical designer, nor Git expert.

geggi:
I have now created icons for 16x16 and 64x64. tried to make them "pixel perfect" (to the degree possible given round shapes) and committed them to the respective folders. If the design is okay, I will also create version for 24x24 (and possibly the other icon sizes as well).

palob:
I'm also only a hobby graphics designer :). If you are not doing this already, you can start using existing ``format-text-*`` icons so (if the **a**s aren't too big) all you need to do is moving the a to the left and adding the superscript/subscript glyphs.

geggi:
Now subscript and superscript icons in format 16, 22, 24, 32, and 64 are available.
I slightly changed the path for the "2" glyphs in small sizes to be better "pixel-perfect".

palob (changes requested):
Thank you so much for doing this.

There's those small issues left. 

+ The 22px/24px icons have some vertical lines misaligned which can be fixed by moving some nodes.

24px

We don't need an extra design for 24 px. The 24px icons usually are the 22px icons with 1px padding around.

+ The *2* glyph in the 16px icons has a small misalignment which can be fixed like this.

16px

+ The 64px superscript icons has the elements misplaced/misaligned.

+ 48px icons are missing. Just the 24px ones scaled up x2 and positioned at coordinates x2.

geggi:
Thank you very much for the additional and detailed guidance.
Actually, I had not known how to edit/move the nodes of the graph, so I first had to find out. It makes absolute sense to align all nodes with the grid, while in the last version I just had aligned the whole graph to the grid.
I hope everything is good now....

PR #1442, opened 2024-03-07
Fix index.theme

vyivel:


palob (commented):
We do actually have one single icon in ``scalable/emotes`` for both Numix and Numix-Light, haven't we?

vyivel:
Yes, you're right. `scalable/emotes`  wasn't listed in `Directories`, however.

Fixed now.

palob (approved):

palob:
Thank you

